PR #2, opened 2023-07-12
Updating copy in "index" and "about" sections

franzihei:
TODO Can somebody please help me fix the links in the ``ContributingCards`` component? Thanks!

### DIFF
--- a/src/components/ContributingCards.tsx
+++ b/src/components/ContributingCards.tsx
@@ -13,24 +13,24 @@ interface ContributingCard {
 }
 const cards: ContributingCard[] = [
   {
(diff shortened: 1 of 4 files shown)


wackerow:
@franzihei pushed a fix... I just ditched the array there and went with regular React.

Just a note, if you use the `Link` component instead of `a`, imported using `import { Link } from '@/components'` it will take care of the styling and internal vs external link handling.

franzihei:
I have no idea what I am doing (never worked with React before), so I just tried to copy the patterns that were there. Thanks for fixing it 😆

PR #7, opened 2023-07-13
Add blog post about missing side effects on selector access bug

r0qs:
Blog post for https://github.com/ethereum/solidity/pull/14366

It was originally created in the old solidity-blog repo: https://github.com/ethereum/solidity-blog/pull/218

### DIFF
--- a/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
+++ b/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
@@ -1,7 +1,7 @@
 ---
 layout: post
 published: true
-title: 'Bug in Legacy Code Generation When Accessing the .selector Member on Expressions with Side-Effects'
+title: 'Bug in Legacy Code Generation When Accessing the .selector Member on Expressions with Side Effects'
 date: '2023-07-19'
 author: Solidity Team
 category: Security Alerts
--- a/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
+++ b/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
@@ -1,0 +1,104 @@
+---
+layout: post
+published: true
+title: 'Bug in Legacy Code Generation When Accessing the .selector Member on Expressions with Side-Effects'
+date: '2023-07-19'
+author: Solidity Team
+category: Security Alerts
+---
+
+On June 26, 2023, a bug in the legacy code generation pipeline of the Solidity compiler was found during
+investigation of a security report related to the use of `abi.decode` with a ternary
+expression that has side effects, as the type argument.
+
+The legacy code generator was not evaluating complex expressions, like assignments, function calls or conditionals,
+whose  `.selector` was being accessed.
+This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
+contracts compiled using the legacy pipeline.
+The via IR code generator behaves as expected and evaluates such expressions.
+
+The bug has been present since version 0.6.2 of the compiler.
+However, to the best of our knowledge, it has never been exploited or reported externally.
+Any actually affected coding pattern would likely be considered an anti-pattern by users and disregarded
+in favor of more sensible alternatives.
+
+For example, in an expression like `f().g.selector`, the result will always be the same regardless
+of what `f()` returns, and it is more natural to perform the operation directly on the contract type,
+i.e., `C.g.selector`.
+Furthermore, expressions like `f().selector` remain unaffected.
+As a result, we have assigned it a severity level of "low".
+
+## Which Contracts Are Affected?
+
+If you are still utilizing the legacy code generation pipeline and your contract accesses the `.selector`
+on expressions that have side effects your code may be vulnerable to this bug.
+For example, the code below, if compiled via IR will result in `x` being
+assigned `42` and contract `D` being created as a side effect of evaluating the function call `h()`.
+On the other hand, if compiled using the legacy code generation pipeline, it will result in `x`
+being `0` and `D` not being deployed.
+
+```solidity
+contract D {
+    function f() external {}
+}
+
+contract C {
+    uint256 x;
+
+    function f() public {
+        h().f.selector;
+    }
+
+    function h() public returns (D) {
+        x = 42;
+        return new D();
+    }
+}
+```
+
+Below, you can find another example that demonstrates the use of an assignment expression instead of a
+function call expression.
+In the example, if the legacy code generation pipeline was used, the expression `x = true`
+would not be evaluated properly, causing the value of `x` to remain unchanged.
+
+```solidity
+contract D {
+    function g() external {}
+}
+
+contract C {
+    bool x;
+    function f() external {
+        ((x = true) ? C.f : D.g).selector;
+    }
+}
+```
+
+## Technical Details
+
+The bug manifested itself in cases where it was possible to determine which function is the result
+of an expression at compilation time, which allowed the compiler to calculate the selector without
+actually evaluating the expression.
+The legacy code generator would take that shortcut to generate more efficient code on the wrong
+assumption that the expression must be side effect free.
+
+In contrast, the new IR-based code generator is primarily concerned with producing correct code and
+can defer the task of making it efficient to the Yul optimizer.
+This means that it does not have to rely on such assumptions.
+The whole expression is always visited and evaluated in the unoptimized code.
+It will only be removed by the optimizer if it is determined to actually be free of side effects.
+
+Due to a related bug in the type checker, the bug would also impact
+certain expressions whose value could not be determined at compilation time.
+The compiler internally associates function types with function definitions, and the associated
+definition would be carried over to the result type of an expression.
+If the expression involved multiple functions, the resulting type would not necessarily point at
+the right one.
+
+For example, even in `(condition ? C.f : D.g).selector`, where the value of `condition` is only
+known at runtime, the compiler was able to determine the common type for `C.f` and `D.g`,
+and would use the definition associated with such type to determine the function to take the selector from.
+This bug was also fixed in the [Solidity v0.8.21](https://github.com/ethereum/solidity/releases/tag/v0.8.21) release.
+
+In conclusion, the fact that this bug remained undiscovered until now indicates that real projects do not seem to be using such
+a pattern since the expected usage of the `.selector` is in a constant expression.
--- a/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
+++ b/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
@@ -11,7 +11,7 @@ On June 26, 2023, a bug in the legacy code generation pipeline of the Solidity c
 investigation of a security report related to the use of `abi.decode` with a ternary
 expression that has side effects, as the type argument.
 
-The legacy code generator was not evaluating complex expressions, like assignments, function calls or conditionals,
+The legacy code generator was not evaluating complex expressions, like assignments, function calls, or conditionals,
 whose  `.selector` was being accessed.
 This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
 contracts compiled using the legacy pipeline.
@@ -86,7 +86,7 @@ In contrast, the new IR-based code generator is primarily concerned with produci
 can defer the task of making it efficient to the Yul optimizer.
 This means that it does not have to rely on such assumptions.
 The whole expression is always visited and evaluated in the unoptimized code.
-It will only be removed by the optimizer if it is determined to actually be free of side effects.
+It will only be removed by the optimizer if it is determined actually to be free of side effects.
 
 Due to a related bug in the type checker, the bug would also impact
 certain expressions whose value could not be determined at compilation time.
--- a/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
+++ b/src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
@@ -15,7 +15,7 @@ The legacy code generator was not evaluating complex expressions, like assignmen
 whose  `.selector` was being accessed.
 This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
 contracts compiled using the legacy pipeline.
-The via IR code generator behaves as expected and evaluates such expressions.
+The via-IR code generator behaves as expected and evaluates such expressions.
 
 The bug has been present since version 0.6.2 of the compiler.
 However, to the best of our knowledge, it has never been exploited or reported externally.


src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,71 @@
...
+The legacy code generator was not evaluating complex expressions, like assignments, function calls or conditionals,
+whose  ``.selector`` was being accessed.
+This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of 
+contracts compiled using the legacy pipeline.
+The via IR code generator behaves as expected and evaluates such expressions.
+
+The bug has been present since the compiler version ``0.6.2``.
+However, to the best of our knowledge, it has never been exploited or reported externally.
+It can be considered an anti-pattern, for example, instead of using ``f().g.selector`` it is more advisable to
+perform the operation directly on the contract, i.e., ``C.g.selector``.
```
cameel, line 23:
```suggestion
Any actually affected coding pattern would likely be considered an anti-pattern by users and disregarded on favor of more sensible alternatives.
For example, in an expression like ``f().g.selector`` the result will always be the same regardless of what ``f()`` returns and it is more natural to perform the operation directly on the contract type, i.e., ``C.g.selector``.
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,71 @@
...
+Furthermore, expressions like ``f().selector`` remain unaffected.
+As a result, we have assigned it a severity level of "low".
+
+## Which Contracts are Affected?
+
+If you are still utilizing the legacy code generation pipeline and your contract accesses the ``.selector``
+on expressions that have side-effects your code may be vulnerable to this bug.
+For example, the code below, if compiled with ``via-ir`` will result in ``x`` being 
+assigned ``42`` and contract ``D`` being created, as a side-effect of evaluating the function call ``h()``.
+On the other hand, if compiled without ``via-ir`` (i.e. using the legacy code generation pipeline)
```
cameel, line 33: I guess you're referring to the CLI option here?
```suggestion
For example, the code below, if compiled with ``--via-ir`` will result in ``x`` being 
assigned ``42`` and contract ``D`` being created, as a side-effect of evaluating the function call ``h()``.
On the other hand, if compiled without ``--via-ir`` (i.e. using the legacy code generation pipeline) 
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,71 @@
...
+whose  ``.selector`` was being accessed.
+This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of 
+contracts compiled using the legacy pipeline.
+The via IR code generator behaves as expected and evaluates such expressions.
+
+The bug has been present since the compiler version ``0.6.2``.
+However, to the best of our knowledge, it has never been exploited or reported externally.
+It can be considered an anti-pattern, for example, instead of using ``f().g.selector`` it is more advisable to
+perform the operation directly on the contract, i.e., ``C.g.selector``.
+Furthermore, expressions like ``f().selector`` remain unaffected.
```
cameel, line 24: I think this still does not clarify very well what kind of expressions are affected. We should either give some more general "rule" to determine this, or at least list more examples if giving such a rule in a way understandable to the user is too hard.

For example function pointers are not affected either. What about built-ins and various kinds or internal/external/library functions?

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
r0qs: I was referring to the compiler feature, actually. Which you can enable using the CLI or the Standard JSON. Maybe it should be something like `...if you are compiling via IR...`.

cameel: Oh. In that case that would be good too. I generally find "via-IR" used as a noun a bit awkward and try to avoid it. Instead I usually say something like "IR-based code generation pipeline" or "IR generator". On the other hand various spellings of "viair" is what we call the options and often use in external communication so it's getting complicated :)

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,101 @@
...
+The bug manifested itself in cases where it was possible to determine which function is the result
+of an expression at compilation time, which allowed the compiler to calculate the selector without
+actually evaluating the expression.
+The legacy code generator would take that shortcut to generate more efficient code, on the wrong
+assumption that the expression must be side-effect free.
+
+In contrast, the new IR-based code generator is primarily concerned with producing correct code and
+can defer the task of making it efficient to the Yul optimizer.
+This means that it does not have to rely on such assumptions.
+The whole expression is always visited and evaluated in the unoptimized code.
```
r0qs, line 41: Isn't the opposite? I.e. the via IR pipeline is not affected by the bug, only the legacy codegen.

```suggestion
The whole expression is always visited and evaluated in the optimized code.
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,101 @@
...
+It will only be removed by the optimizer if it is determined to actually be free of side-effects.
+
+Due to a related bug in the type checker, the bug would also impact
+certain expressions whose value could not be determined at compilation time.
+The compiler internally associates function types with function definitions, and values of the same
+function type pointing at different definitions were implicitly convertible to each other,
+resulting in the associated definition being carried over to the resulting type.
```
cameel, line 48: Like I said on the chat, this is not really due to the implicit conversion. That affected cases where the expression is actually evaluated. There the problems is just due to the propagation of the definition. Let's reword it like this:

```suggestion
The compiler internally associates function types with function definitions, and the associated
definition would be carried over to the result type of an expression.
If the expression involved multiple functions, the resulting type would not necessarily point at
the right one.
```

cameel (commented):
The post seems fine overall.

There a few minor comments still unaddressed but the content itself is already good enough.

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,101 @@
...
+For example, even in ``(condition ? C.f : D.f).selector``, where the value of ``condition`` is only
+known at runtime, the compiler was able to determine the common type for ``C.f`` and ``D.f``,
```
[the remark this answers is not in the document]
r0qs: Oh, but wasn't this disallowed by the restrict mobile types PR?

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,102 @@
...
+    bool x;
+
+    function f() public {
+        ((x = true) ? D.f : D.f).selector;
+    }
+}
+```
+
+However, the fact that this bug remained undiscovered until now indicates that real projects do not seem to be using such
+a pattern, since the expected usage of the `.selector` is in a constant expression.
```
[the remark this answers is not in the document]
r0qs: Right, it was supposed to be a sort of conclusion.

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+
+On June 26, 2023, a bug in the legacy code generation pipeline of the Solidity compiler was found during
+investigation of a security report related to the use of `abi.decode` with a ternary
+expression, that has side-effects, as the type argument.
+
+The legacy code generator was not evaluating complex expressions, like assignments, function calls or conditionals,
+whose  `.selector` was being accessed.
+This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
+contracts compiled using the legacy pipeline.
+The via IR code generator behaves as expected and evaluates such expressions.
```
NunoFilipeSantos, line 18:
```suggestion
The via-IR code generator behaves as expected and evaluates such expressions.
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+
+The bug has been present since version 0.6.2 of the compiler.
+However, to the best of our knowledge, it has never been exploited or reported externally.
+Any actually affected coding pattern would likely be considered an anti-pattern by users and disregarded
+in favor of more sensible alternatives.
+For example, in an expression like `f().g.selector` the result will always be the same regardless
+of what `f()` returns and it is more natural to perform the operation directly on the contract type,
+i.e., `C.g.selector`.
+Furthermore, expressions like `f().selector` remain unaffected.
+As a result, we have assigned it a severity level of "low".
```
NunoFilipeSantos, line 28:
```suggestion
For example, in an expression like `f().g.selector`, the result will always be the same regardless
of what `f()` returns, and it is more natural to perform the operation directly on the contract type,
i.e., `C.g.selector`.
Furthermore, expressions like `f().selector` remain unaffected.
As a result, we have assigned it a severity level of "low".
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Technical Details
+
+The bug manifested itself in cases where it was possible to determine which function is the result
+of an expression at compilation time, which allowed the compiler to calculate the selector without
+actually evaluating the expression.
+The legacy code generator would take that shortcut to generate more efficient code, on the wrong
+assumption that the expression must be side-effect free.
```
NunoFilipeSantos, line 36:
```suggestion
assumption that the expression must be side effect free.
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+i.e., `C.g.selector`.
+Furthermore, expressions like `f().selector` remain unaffected.
+As a result, we have assigned it a severity level of "low".
+
+## Technical Details
+
+The bug manifested itself in cases where it was possible to determine which function is the result
+of an expression at compilation time, which allowed the compiler to calculate the selector without
+actually evaluating the expression.
+The legacy code generator would take that shortcut to generate more efficient code, on the wrong
```
NunoFilipeSantos, line 35:
```suggestion
The legacy code generator would take that shortcut to generate more efficient code on the wrong
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+assumption that the expression must be side-effect free.
+
+In contrast, the new IR-based code generator is primarily concerned with producing correct code and
+can defer the task of making it efficient to the Yul optimizer.
+This means that it does not have to rely on such assumptions.
+The whole expression is always visited and evaluated in the unoptimized code.
+It will only be removed by the optimizer if it is determined to actually be free of side-effects.
```
NunoFilipeSantos, line 42:
```suggestion
It will only be removed by the optimizer if it is determined to actually be free of side effects.
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+
+For example, even in `(condition ? C.f : D.g).selector`, where the value of `condition` is only
+known at runtime, the compiler was able to determine the common type for `C.f` and `D.g`,
+and would use the definition associated with such type to determine the function to take the selector from.
+This bug was also fixed in this release.
+
+## Which Contracts Are Affected?
+
+If you are still utilizing the legacy code generation pipeline and your contract accesses the `.selector`
+on expressions that have side-effects your code may be vulnerable to this bug.
```
NunoFilipeSantos, line 59:
```suggestion
on expressions that have side effects your code may be vulnerable to this bug.
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+For example, the code below, if compiled via IR will result in `x` being
+assigned `42` and contract `D` being created, as a side-effect of evaluating the function call `h()`.
```
NunoFilipeSantos, line 61:
```suggestion
assigned `42` and contract `D` being created as a side effect of evaluating the function call `h()`.
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+For example, even in `(condition ? C.f : D.g).selector`, where the value of `condition` is only
+known at runtime, the compiler was able to determine the common type for `C.f` and `D.g`,
+and would use the definition associated with such type to determine the function to take the selector from.
+This bug was also fixed in this release.
+
+## Which Contracts Are Affected?
+
+If you are still utilizing the legacy code generation pipeline and your contract accesses the `.selector`
+on expressions that have side-effects your code may be vulnerable to this bug.
+For example, the code below, if compiled via IR will result in `x` being
```
NunoFilipeSantos, line 60:
```suggestion
For example, the code below, if compiled with via-IR will result in `x` being
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+assigned `42` and contract `D` being created, as a side-effect of evaluating the function call `h()`.
+On the other hand, if compiled using the legacy code generation pipeline it will result in `x`
```
NunoFilipeSantos, line 62:
```suggestion
On the other hand, if compiled using the legacy code generation pipeline, it will result in `x`
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+    bool x;
+    function f() external {
+        ((x = true) ? C.f : D.g).selector;
+    }
+}
+```
+
+
+In conclusion, the fact that this bug remained undiscovered until now indicates that real projects do not seem to be using such
+a pattern, since the expected usage of the `.selector` is in a constant expression.
```
NunoFilipeSantos, line 104:
```suggestion
a pattern since the expected usage of the `.selector` is in a constant expression.
```

NunoFilipeSantos (changes requested):
I left a few notes.

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
r0qs: See https://github.com/ethereum/solidity-website/pull/7#discussion_r1265609231 for the reasoning of why we decided to go for `via IR` instead of `via-IR`

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
r0qs: See https://github.com/ethereum/solidity-website/pull/7#discussion_r1267791812

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+published: true
+title: 'Bug in Legacy Code Generation When Accessing the .selector Member on Expressions with Side-Effects'
+date: '2023-07-19'
+author: Solidity Team
+category: Security Alerts
+---
+
+On June 26, 2023, a bug in the legacy code generation pipeline of the Solidity compiler was found during
+investigation of a security report related to the use of `abi.decode` with a ternary
+expression, that has side-effects, as the type argument.
```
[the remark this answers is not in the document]
r0qs: I agree with the comma suggestion, but why not `side-effects`?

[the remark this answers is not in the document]
cameel: I looked that up and people are using both versions, but apparently the one without hyphen is vastly more common.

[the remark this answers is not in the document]
NunoFilipeSantos: https://www.merriam-webster.com/dictionary/side%20effect

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+
+The bug has been present since version 0.6.2 of the compiler.
+However, to the best of our knowledge, it has never been exploited or reported externally.
+Any actually affected coding pattern would likely be considered an anti-pattern by users and disregarded
+in favor of more sensible alternatives.
+For example, in an expression like `f().g.selector`, the result will always be the same regardless
+of what `f()` returns, and it is more natural to perform the operation directly on the contract type,
+i.e., `C.g.selector`.
+Furthermore, expressions like `f().selector` remain unaffected.
+As a result, we have assigned it a severity level of "low".
```
NunoFilipeSantos, line 28:
```suggestion
On June 26, 2023, a bug in the legacy code generation pipeline of the Solidity compiler was found during
investigation of a security report related to the use of `abi.decode` with a ternary
expression that has side effects, as the type argument.
The bug has been present since version 0.6.2 of the compiler.
However, to the best of our knowledge, it has never been exploited or reported externally.
Any actually affected coding pattern would likely be considered an anti-pattern by users and disregarded
in favor of more sensible alternatives.

The legacy code generator was not evaluating complex expressions, like assignments, function calls or conditionals,
whose  `.selector` was being accessed.
This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
contracts compiled using the legacy pipeline.
The via IR code generator behaves as expected and evaluates such expressions.

For example, in an expression like `f().g.selector`, the result will always be the same regardless
of what `f()` returns, and it is more natural to perform the operation directly on the contract type,
i.e., `C.g.selector`.
Furthermore, expressions like `f().selector` remain unaffected.
As a result, we have assigned it a severity level of "low".
```

NunoFilipeSantos: @r0qs this way, a user doesn't have to jump between paragraphs to understand wince when it has been present before going into tech details.

NunoFilipeSantos: This way, it also matches @cameel's blogpost structure.

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+    function g() external {}
+}
+
+contract C {
+    bool x;
+    function f() external {
+        ((x = true) ? C.f : D.g).selector;
+    }
+}
+```
```
NunoFilipeSantos, line 100:
```suggestion
## Which Contracts Are Affected?

If you are still utilizing the legacy code generation pipeline and your contract accesses the `.selector`
on expressions that have side effects your code may be vulnerable to this bug.
For example, the code below, if compiled via IR will result in `x` being
assigned `42` and contract `D` being created as a side effect of evaluating the function call `h()`.
On the other hand, if compiled using the legacy code generation pipeline, it will result in `x`
being `0` and `D` not being deployed.

```solidity
contract D {
    function f() external {}
}

contract C {
    uint256 x;

    function f() public {
        h().f.selector;
    }

    function h() public returns (D) {
        x = 42;
        return new D();
    }
}
```

Below, you can find another example that demonstrates the use of an assignment expression instead of a
function call expression.
In the example, if the legacy code generation pipeline was used, the expression `x = true`
would not be evaluated properly, causing the value of `x` to remain unchanged.

```solidity
contract D {
    function g() external {}
}

contract C {
    bool x;
    function f() external {
        ((x = true) ? C.f : D.g).selector;
    }
}
```
## Technical Details

The bug manifested itself in cases where it was possible to determine which function is the result
of an expression at compilation time, which allowed the compiler to calculate the selector without
actually evaluating the expression.
The legacy code generator would take that shortcut to generate more efficient code on the wrong
assumption that the expression must be side effect free.

In contrast, the new IR-based code generator is primarily concerned with producing correct code and
can defer the task of making it efficient to the Yul optimizer.
This means that it does not have to rely on such assumptions.
The whole expression is always visited and evaluated in the unoptimized code.
It will only be removed by the optimizer if it is determined to actually be free of side effects.

Due to a related bug in the type checker, the bug would also impact
certain expressions whose value could not be determined at compilation time.
The compiler internally associates function types with function definitions, and the associated
definition would be carried over to the result type of an expression.
If the expression involved multiple functions, the resulting type would not necessarily point at
the right one.

For example, even in `(condition ? C.f : D.g).selector`, where the value of `condition` is only
known at runtime, the compiler was able to determine the common type for `C.f` and `D.g`,
and would use the definition associated with such type to determine the function to take the selector from.
This bug was also fixed in this release.
```

File: src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md [hunk 21]
```diff
@@ -0,0 +1,104 @@
...
+certain expressions whose value could not be determined at compilation time.
+The compiler internally associates function types with function definitions, and the associated
+definition would be carried over to the result type of an expression.
+If the expression involved multiple functions, the resulting type would not necessarily point at
+the right one.
+
+For example, even in `(condition ? C.f : D.g).selector`, where the value of `condition` is only
+known at runtime, the compiler was able to determine the common type for `C.f` and `D.g`,
+and would use the definition associated with such type to determine the function to take the selector from.
+This bug was also fixed in this release.
```
NunoFilipeSantos, line 54:
```suggestion
## Which Contracts Are Affected?

If you are still utilizing the legacy code generation pipeline and your contract accesses the `.selector`
on expressions that have side effects your code may be vulnerable to this bug.
For example, the code below, if compiled via IR will result in `x` being
assigned `42` and contract `D` being created as a side effect of evaluating the function call `h()`.
On the other hand, if compiled using the legacy code generation pipeline, it will result in `x`
being `0` and `D` not being deployed.

```solidity
contract D {
    function f() external {}
}
contract C {
    uint256 x;
    function f() public {
        h().f.selector;
    }
    function h() public returns (D) {
        x = 42;
        return new D();
    }
}
```

Below, you can find another example that demonstrates the use of an assignment expression instead of a
function call expression.
In the example, if the legacy code generation pipeline was used, the expression `x = true`
would not be evaluated properly, causing the value of `x` to remain unchanged.

```solidity
contract D {
    function g() external {}
}
contract C {
    bool x;
    function f() external {
        ((x = true) ? C.f : D.g).selector;
    }
}
```
```

File: src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md (hunk 20 above)
Comment (NunoFilipeSantos) [inline src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md:100]: Please swap these two sections to match @cameel 's structure.
Users want to know first Which contracts are affected to know they're safe, not the technical details.

File: src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md (hunk 19 above)
Comment (r0qs) [reply src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md]: Oh, actually, I think it was better the way it was before, because that explanation of the bug was necessary, in my point of view, to understand why we classify the severity level as low.

Comment (NunoFilipeSantos) [reply src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md]: The explanation is still there, and so it is the "low" classification in the last sentence.

Or are you suggesting to just move to the bottom the following text? If so, I'm cool.
```
The legacy code generator was not evaluating complex expressions, like assignments, function calls or conditionals,
whose  `.selector` was being accessed.
This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
contracts compiled using the legacy pipeline.
The via IR code generator behaves as expected and evaluates such expressions.
```

Comment (r0qs) [reply src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md]: This:
```
On June 26, 2023, a bug in the legacy code generation pipeline of the Solidity compiler was found during
investigation of a security report related to the use of `abi.decode` with a ternary
expression that has side effects, as the type argument.

The legacy code generator was not evaluating complex expressions, like assignments, function calls or conditionals,
whose  `.selector` was being accessed.
This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
contracts compiled using the legacy pipeline.
The via IR code generator behaves as expected and evaluates such expressions.

For example, in an expression like `f().g.selector`, the result will always be the same regardless
of what `f()` returns, and it is more natural to perform the operation directly on the contract type,
i.e., `C.g.selector`.
Furthermore, expressions like `f().selector` remain unaffected.

The bug has been present since version 0.6.2 of the compiler.
However, to the best of our knowledge, it has never been exploited or reported externally.
Any actually affected coding pattern would likely be considered an anti-pattern by users and disregarded
in favor of more sensible alternatives.
As a result, we have assigned it a severity level of "low".

```

File: src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md (hunk 8 above)
Comment (NunoFilipeSantos) [reply src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md:18]: [Developer Survey](https://soliditylang.org/blog/2023/03/10/solidity-developer-survey-2022-results/): via-IR
[Documentation](https://docs.soliditylang.org/en/v0.8.21/ir-breaking-changes.html): via-ir.

Let's not add yet another confusing term.

File: src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md [hunk 22]
```diff
@@ -0,0 +1,104 @@
+---
+layout: post
+published: true
+title: 'Bug in Legacy Code Generation When Accessing the .selector Member on Expressions with Side-Effects'
```
NunoFilipeSantos, line 4:
```suggestion
title: 'Bug in Legacy Code Generation When Accessing the .selector Member on Expressions with Side Effects'
```

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -15,7 +15,7 @@ The legacy code generator was not evaluating complex expressions, like assignmen
 whose  `.selector` was being accessed.
 This led to the side-effects of such expressions not being executed, and therefore potentially incorrect behavior of
 contracts compiled using the legacy pipeline.
-The via IR code generator behaves as expected and evaluates such expressions.
+The via-IR code generator behaves as expected and evaluates such expressions.
```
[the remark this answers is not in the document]
NunoFilipeSantos: [Developer Survey](https://soliditylang.org/blog/2023/03/10/solidity-developer-survey-2022-results/): via-IR
[Documentation](https://docs.soliditylang.org/en/v0.8.21/ir-breaking-changes.html): via-ir.

Let's not add yet another confusing term.

NunoFilipeSantos (approved):
Ship it.

src/posts/2023-07-19-missing-side-effects-on-selector-access-bug.md
```diff
@@ -0,0 +1,104 @@
...
+If the expression involved multiple functions, the resulting type would not necessarily point at
+the right one.
+
+For example, even in `(condition ? C.f : D.g).selector`, where the value of `condition` is only
+known at runtime, the compiler was able to determine the common type for `C.f` and `D.g`,
+and would use the definition associated with such type to determine the function to take the selector from.
+This bug was also fixed in the [Solidity v0.8.21](https://github.com/ethereum/solidity/releases/tag/v0.8.21) release.
+
+In conclusion, the fact that this bug remained undiscovered until now indicates that real projects do not seem to be using such
+a pattern since the expected usage of the `.selector` is in a constant expression.
```
cameel, line 104: TBH, this sentence still feels like it comes out of nowhere and does not follow logically from the result of the section :)

cameel (approved):
I have a small nitpick, but nothing major, so we could also just post it as is.

PR #12, opened 2023-07-17
Change link for Solidity Summit 2023 to Devconnect page

franzihei:
We'll direct to https://devconnect.org/schedule until we have updated summit.soliditylang.org


NunoFilipeSantos (approved):
Thank you!

PR #20, opened 2023-07-20
Use internal link for blog callout

minimalsm:
Our homepage currently uses the 'blog.soliditylang.org' link. This is recognized as an external link, which triggers a complete page refresh and opens in a new tab.

In this PR, I've updated the link to use 'soliditylang.org/blog/', the recognized internal link. This should avoid the page reloading and provide a smoother user experience.


wackerow (approved):
Thanks @minimalsm!

PR #33, opened 2023-07-24
Update Solidity Summit 2023 Card to point to summit.soliditylang.org

franzihei:
I updated https://summit.soliditylang.org/ to a 2023 "coming soon" stage so I believe we can add the link back into the main page. :)


NunoFilipeSantos (approved):
LFG!

PR #38, opened 2023-08-04
Events path to root and latest Summit path rename

wackerow:
## Description

- Moves all events to root level paths
- Renames the current/latest Solidity Summit event to simply `summit.md` to enable the use of `soliditylang.org/summit`
- In subsequent years, this can be renamed to something like `solidity-summit-2023.md` for archival purposes, and replaced with a new `summit.md` for the next year.

With this we can readdress https://github.com/ethereum/solidity-website/pull/36#discussion_r1284599046 using `https://soliditylang.org/summit/`, and also update the `https://summit.soliditylang.org` to point to this as well.

cc @franzihei @NunoFilipeSantos


franzihei:
Great! We still need to update https://summit.soliditylang.org to point to https://soliditylang.org/summit/, correct? I guess this needs to be done by devops since it's domain related?

NunoFilipeSantos:
Yes, it's a ´301´redirect.

wackerow:
Ah, that was meant as a general "we"... I do not have access here, but on that note I am more than happy to ping devops on this... I will file a ticket and ping in Discord.

PR #46, opened 2023-08-27
Force non-emoji unicode arrow [Fixes #6]

### LINKED ISSUE #6: [Bug] Inconsistent Display of Up-and-Right Arrow on Certain Android Devices
Requires more investigation:

On certain Android devices, the text arrow (`This is a link ↗`) is displayed as an emoji arrow (This is a link ↗️) instead. The expected behaviour is for the arrow to always appear as text (`This is a link ↗`). The cause, the exact Android versions, and the devices affected need to be identified. 

Community input and suggestions for potential solutions are welcome ❤️ 

Screenshot 2023-07-13 at 11 25 51

wackerow:
## Description
Adds unicode modifier to force the external link arrow to always display in the non-emoji force.

The modifier code `FE0E` (`\uFE0E` in JavaScript) can be placed after a unicode character to avoid ambiguity and force the non-emoji rendering:

image

---

The opposite can be done with `\uFE0F` which will force the emoji form:

image

## Related issue
- Fixes #6

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -35,7 +35,7 @@ export const Link: React.FC<LinkComponentProps> = ({
           hideArrow
             ? {}
             : {
-                content: '" ↗"',
+                content: '" ↗\uFE0E"',
                 whiteSpace: 'nowrap',
               }
         }


r0qs (approved):
I'm approving based on the explanation of the solution here https://github.com/ethereum/solidity-website/issues/6#issuecomment-1694769052. However, I couldn't reproduce the problem on some devices I tested. So it would be nice if @minimalsm could inform us what Android version are affected.

PR #57, opened 2023-11-02
Post describing the verbatim invalid deduplication bug

matheusaaguiar:


src/posts/2023-11-7-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,58 @@
...
+  }          
+
+  sstore(0, output)
+}
+```
+
+In this case, the block deduplicator incorrectly considers that the blocks
+of cases `0x00` and `0x02` are identical and thus can be merged into a single one.
+
+## Technical Details
```
matheusaaguiar, line 58: Not sure if we need this section for this bug. @ekpyron @cameel

cameel: I think it's always a good idea to give full details, so that people can try to understand and evaluate it for themselves. For example, currently the post does not really make it clear that the deduplicated blocks are really blocks of EVM assembly instructions and not Yul blocks. We could use some background here, with links to the docs.

src/posts/2023-11-7-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,58 @@
...
+`verbatim` items are surrounded by identical opcodes, they are considered equal
+and thus could be incorrectly unified.
+The bug exists since version `0.8.5` and does **not** affect Solidity code,
+only Yul code. It can be avoided by turning off the flag `deduplicate` in the
+optimizer details settings in the via the [standard-json interface](https://docs.soliditylang.org/en/v0.8.21/using-the-compiler.html#input-description).
+Although the use of `verbatim` is not much uncommon the conditions which trigger
+the bug, two or more `verbatim` items surrounded by identical opcodes, are very specific.
+Also, to the extent of the investigations made by the team, there is no evidence
+that this could be used as an exploit or attack vector.
+Considering that, the team assigned the bug a severity level of `low`.
```
cameel, line 24: We should rate severity and likelihood separately.

src/posts/2023-11-7-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,125 @@
...
+The block deduplicator identifies the start of a block upon finding a tag. The
+rest of the block is then formed by the sequence of assembly opcodes until the
+first `JUMP` or other opcode capable of altering the control flow, such as
+`RETURN` or REVERT.
+When the optimizer encounters a tag id `Ta`, it identifies the beginning of a
+block of items starting at a certain position `i`. Then, the optimizer checks
+if there is a previously seen block of identical assembly items, which starts at
+position `j` and is referenced by a tag `Tb`. In case such a block exists, the
+optimizer maps `Tb` as a replacement for the tag `Ta`, which means that later
+one block will be not referred to by any tag and can then be erased.
```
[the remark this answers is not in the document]
matheusaaguiar: I am not sure. I just assumed that, but didn't check.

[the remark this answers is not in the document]
cameel: It's also fine to just reword this to be more neutral. Like, say that it will not be included in the final bytecode.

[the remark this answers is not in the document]
ekpyron: > I guess it's more that it will not be included in the output when the bytecode is reconstructed from the control flow graph rather than actively erased from it? Or is it?

@matheusaaguiar's description is completely correct - the tag is replaced on the jump, which leads to one of the tags becoming unused and then the unused tag is erased by the ``JumpdestRemover`` :-).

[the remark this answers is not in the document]
ekpyron: Note that technically for the bug the removal of the blog doesn't really matter - even if you disable the jumpdest remover, the blog actually remains, but the jumps that were supposed to jump to it jump to a different block due to the tag being replaced by the deduplicator.

Not entirely sure if we're not actually going a bit overboard in the description here :-). But also doesn't hurt.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,188 @@
...
+previously seen block and the next one is found and processed.
+
+The bug manifested when checking if a given block `i` is identical to a
+previously seen block `j`. The comparison between two assembly blocks is based
+on the [operator `<`](https://github.com/ethereum/solidity/blob/v0.8.22/libevmasm/AssemblyItem.h#L136-L146)
+of assembly item. The bug only affected `verbatim` items, as their bytecode
+contents were incorrectly compared using operator `==`, instead of operator `<`.
+
+For example, the Yul code snippet showed before, when compiled **without**
+optimizations will include the assembly code fragment shown below:
```
mehtavishwa30, line 105:
```suggestion
For example, when the Yul code snippet shown before is compiled **without**
optimizations, it will include the assembly code fragment shown below:
```

small nitpick. does it read better?

cameel:
I tweaked the wording in the initial part of post a bit. Done as fixups to avoid spamming you with comments. Please take a look and let me know if there's something that you don't like.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+`verbatim` assembly items surrounded by identical opcodes were incorrectly considered identical
+and unified.
+
+The bug existed since version `0.8.5`, which introduced `verbatim`, and only affected pure Yul
+compilation with optimization enabled.
+Solidity code or Yul used in inline assembly blocks would not trigger it.
+To mitigate it, it is sufficient to turn off the `settings.optimizer.details.deduplicate` flag in the
+in [Standard JSON input](https://docs.soliditylang.org/en/v0.8.23/using-the-compiler.html#input-description).
+
+Although the use of `verbatim` is not uncommon, the conditions which trigger
```
ekpyron, line 24: Yeah? I thought we concluded that nobody actually uses `verbatim` at all, didn't we :-)?

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+the bug (two or more `verbatim` items surrounded by identical opcodes) are very specific.
+Also, to the extent of the investigations made by the team, there is no evidence
+that this could be used as an exploit or attack vector.
+Considering that, the team assigned the bug a severity level of `low`.
+
+## Which Contracts Are Affected?
+
+The precise conditions which triggered the bug are as follows:
```
ekpyron, line 32: The list below doesn't have "precise conditions", but "necessary conditions" - if we state those as precise decisions, it makes it sound like *any* use of verbatim is affected, while in fact it's still highly unlikely that you hit it if the conditions here are true (you need very precise patterns to achieve fully identical blocks on the evm assembly level - identical there actually means identical *including* outgoing jumps - that can almost only happen in switches and at the very end of functions to begin with)

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+
+1. Compilation of pure Yul code.
+1. The use of the `verbatim` builtin.
+1. Block Deduplicator optimizer step being in use.
+
+Note that Block Deduplicator is enabled by default when the optimizer is enabled.
+In Standard JSON input it can be explicitly disabled using the `deduplicate` setting.
+On the command line, it can only be disabled along with the whole opcode-based optimizer.
```
ekpyron, line 40: We should not talk about disabling it, that's not a sane thing to do for anyone. Mentioning that it's enabled by default is fine, but let's drop the part about disabling it.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+
+If your project does not include contracts written purely in Yul,
```
ekpyron, line 42:
```suggestion
If your project does not include contracts written purely in Yul or does not use ``verbatim``,
```

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+The block deduplicator identifies the start of a block upon finding a tag. The
+rest of the block is then formed by the sequence of assembly opcodes until the
+first `JUMP` or other opcode capable of altering the control flow, such as
+`RETURN` or `REVERT`.
+When the optimizer encounters a tag id `Ta`, it identifies the beginning of a
+block of items starting at a certain position `i`. Then, the optimizer checks
+if there is a previously seen block of identical assembly items, which starts at
+position `j` and is referenced by a tag `Tb`. In case such a block exists, the
+optimizer maps `Tb` as a replacement for the tag `Ta`, which means that later
+one block will be not referred to by any tag and can then be erased.
```
matheusaaguiar, line 94:
```suggestion
one block will not be referred to by any tag and can then be erased.
```
I guess this is the correct phrasing.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+If no such block exists, the block referred to by `Ta` gets registered as a
+previously seen block and the next one is found and processed.
+
+The bug manifested when checking if a given block `i` is identical to a
+previously seen block `j`. The comparison between two assembly blocks is based
+on the [operator `<`](https://github.com/ethereum/solidity/blob/v0.8.22/libevmasm/AssemblyItem.h#L136-L146)
+of assembly item. The bug only affected `verbatim` items, as their bytecode
+contents were incorrectly compared using operator `==`, instead of operator `<`.
```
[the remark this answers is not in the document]
ekpyron: If we want, we can describe the effect instead, which is that for the block deduplicator all verbatim items are equal, that's enough.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+  pop
+  jump(tag_1)
+```
+
+Notice that the blocks referred to by `tag_2` and `tag_4` are almost identical.
+If such code would be optimized, when the deduplicator looks at the sequence of
+both blocks, due to the bug, the `verbatim` items were considered equal, despite
+their different contents. That happened because, as explained before, the contents
+should be compared using the operator `<`, but for `verbatim` items, that was
+incorrectly implemented by comparing their contents using operator `==`.
```
ekpyron, line 136: Again, that's too much technical details about the comparison here.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+tag_4:
+  dup4
+  verbatimbytecode_506002
+  swap2
+  pop
+  jump(tag_1)
+```
+
+Notice that the blocks referred to by `tag_2` and `tag_4` are almost identical.
+If such code would be optimized, when the deduplicator looks at the sequence of
```
matheusaaguiar, line 132:
```suggestion
If such code would be optimized, when the deduplicator looked at the sequence of
```

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,188 @@
...
+their different contents. That happened because, as explained before, the contents
+should be compared using the operator `<`, but for `verbatim` items, that was
+incorrectly implemented by comparing their contents using operator `==`.
+So, when comparing the contents of the `verbatim` items of the example,
+`506000` < `506002` was `false` as was `506002` < `506000`. In this way,
+the contents were concluded to be equal and thus, the optimizer would map `tag_4`
+as a replacement for `tag_2`, which was clearly a mistake. That would cause the
+block referred to by `tag_2` to be erased later. The result of the bug can be
+seen in the output of the optimized code next. Notice that only the block with
+`verbatimbytecode_506002` instruction is left now:
```
mehtavishwa30, line 143: Rewritten to suggest some changes in the sentence structure. Feel free to add or ignore.

```suggestion
As a result of the bug, the deduplication step of the optimizer would consider the `verbatim` items as equals despite their different contents. Note that the `<` operator should be used to make the comparison between the contents of the items. The incorrect comparison of the contents occurs due to the usage of the `==` operator instead of the `<` operator to make the comparison.

You can see that in the example, the comparisons of the `verbatim` items `506000` < `506002` and `506002` < `506000` were false and hence, the contents were concluded to be equal. The optimizer would then _incorrectly_ map `tag_4`
as a replacement for `tag_2` which would further cause the
block referred to by `tag_2` to be erased later.

The result of the bug can be
seen in the output of the optimized code below:

```

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
matheusaaguiar: True :sweat_smile: 
I will rewrite saying that the use of `verbatim` **is** uncommon and the conditions which trigger it are very specific.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,188 @@
...
+  sstore
+  stop
+tag_3:
+  pop
+  pop
+  0x01
+  0x00
+  sstore
+  stop
+```
```
mehtavishwa30, line 195: Maybe add the following sentence here instead?

> Notice that only the block with
`verbatimbytecode_506002` instruction remains in the output shown above as a result of the bug.

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,195 @@
...
+one block will be not referred to by any tag and can then be erased.
+If no such block exists, the block referred to by `Ta` gets registered as a
+previously seen block and the next one is found and processed.
+
+The bug manifested when checking if a given block `i` is identical to a
+previously seen block `j`. The comparison between two assembly blocks is based
+on the [operator `<`](https://github.com/ethereum/solidity/blob/v0.8.22/libevmasm/AssemblyItem.h#L136-L146)
+of assembly item. The bug only affected `verbatim` items, as their bytecode
+contents were incorrectly compared using operator `==`, instead of operator `<`.
+
```
ekpyron, line 103: I'm generally not happy about this description.
This is a very convoluted way of saying it compares two blocks by comparing all assembly items in them in sequence. There is no benefit in following the implementation by tracing indices here, that only makes it harder to understand what's happening.

src/posts/2023-11-7-verbatim-invalid-deduplication-bug.md
cameel: I'd say the likelihood is very low. Otherwise it would not get low overall score because severity in affected cases is high.

The situation here is pretty much identical to what we had back in [FullInliner Non-Expression-Split Argument Evaluation Order Bug](https://soliditylang.org/blog/2023/07/19/full-inliner-non-expression-split-argument-evaluation-order-bug/) and that's how I described it back then:
> We assigned the bug an overall score of "low". The bug has "high" severity in affected cases, but we deem the likelihood of it actually affecting deployed contracts as "very low".

src/posts/2023-11-08-verbatim-invalid-deduplication-bug.md
```diff
@@ -0,0 +1,188 @@
+---
+layout: post
+published: true
+title: 'Bug in Block Deduplicator Considering Verbatim Blocks Identical'
+date: '2023-11-08'
+author: Solidity Team
+category: Security Alerts
+---
+
+On October 24, Ori Pomerantz reported a bug regarding the use of `verbatim`
```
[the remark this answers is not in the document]
cameel: Well, it probably makes sense, but just wanted to note that we almost never do that in those blog posts, not sure why.

cameel (approved):

PR #76, opened 2024-01-16
Transient storage blog post.

ekpyron:
Blog post draft relating to https://github.com/ethereum/solidity/pull/14737


src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,92 @@
...
+
+We recommend to generally always clear transient storage completely at the end of a call into your smart contract to avoid these kinds of issues and to simplify the analysis of the behaviour of your contract within complex transactions. As a matter of fact, the Solidity team has been advocating for a change in the specification of transient storage to change its scope to the call frame into a smart contract within a transaction to avoid this kind of pitfall on the EVM level - however, the concern was ultimately ignored and, consequently, responsible and safe use of transient storage now lies within the responsibility of the user. We're still investigating our options to mitigate these pitfalls in future high-level language constructs built on top of the base functionality of the transient storage opcodes.
```
cameel, line 88: I think that "change its scope to the call frame into a smart contract within a transaction" may require some clarification. I originally assumed it meant always resetting transient storage when exiting the current call frame, but that would not make sense. I found [Chris' answer that explains it in the orignal EIP-1153 thread](https://ethereum-magicians.org/t/eip-1153-transient-storage-opcodes/553/73), so, for reference:

> For example, if you have a reentrancy lock and people do not reset it (because they want to save the gas) and rely on it being reset at the end of the transaction, you will no longer be able to call this contract for the rest of the transaction at all.
>
> We can solve this problem by resetting the transient storage of a contract whenever a call returns from the contract and the same contract is not in the current call stack. It would eliminate some other use cases, but it would feel much cleaner to me.

We should also add an example here, showing how it works now and how it would work if transient storage was implemented this way.

cameel (commented):
I have some comments but mostly as suggestions to whoever will be expanding this into a full article.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,92 @@
...
+published: true
+title: 'Transient Storage Opcodes in Solidity 0.8.24'
+date: '2024-01-23'
+author: Solidity Team
+category: Explainers
+---
+
+TODO: adjust date and links!
+
+[Solidity 0.8.24](https://TODO/link/to/release/post) supports the opcodes included in the upcoming Cancun hardfork and, in particular, the transient storage opcodes `TSTORE` and `TLOAD` [as per EIP-1153](https://eips.ethereum.org/EIPS/eip-1153).
```
cameel, line 12: Perhaps we should mention that the optimizer support for transient storage is still very rudimentary. Opcodes are usable and won't break anything, but their use not necessarily as well optimized as for storage and memory. For example the unused store remover won't touch transient storage.

ekpyron: I don't think that's worth mentioning. Without high-level language support, the expected use is fine-tuned anyways and optimizations won't matter.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,92 @@
...
+
+For smart contracts in particular, it is important that their behaviour is self-contained in this manner, such that multiple calls into individual smart contracts can be composed to more complex applications. So far the EVM largely guaranteed composable behaviour, since multiple calls into a smart contract within a complex transaction are virtually indistinguishable from multiple calls to the contract stretched over several transactions (TODO: maybe mention the distinction between tx.origin and msg.sender and related bugs). However, transient storage allows to violate this principle and incorrect use may lead to complex bugs that only surface in use across several calls.
+
+For example, since transient storage is constructed as a relatively cheap key-value store, a smart contract author may be tempted to use transient storage as a replacement for in-memory mappings without keeping track of the modified keys in the mapping and thereby without clearing the mapping at the end of the call. This, however, can easily lead to unexpected behaviour in complex transactions, in which values set by a previous call into the contract within the same transaction remain.
```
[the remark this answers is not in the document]
ekpyron: I actually wouldn't add an example here. The entire point is that you *shouldn't* do this, so I wouldn't provide examples for people to get them bad ideas about things they should *not* do.

And well, in-memory mappings are a recurring feature request that always requires explaining that the way memory works we can't implement cheap hash maps as in storage. But yeah, it happens that a local associative container would be handy for whatever reason.

src/posts/2024-01-23-transient-storage.md
ekpyron: Yep - I actually usually said "outermost call frame", not sure why I didn't here - but maybe best to reproduce Chris's explanation in full detail.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,92 @@
...
+date: '2024-01-23'
+author: Solidity Team
+category: Explainers
+---
+
+TODO: adjust date and links!
+
+[Solidity 0.8.24](https://TODO/link/to/release/post) supports the opcodes included in the upcoming Cancun hardfork and, in particular, the transient storage opcodes `TSTORE` and `TLOAD` [as per EIP-1153](https://eips.ethereum.org/EIPS/eip-1153).
+
+Transient storage is a long-awaited feature on the EVM level that introduces another data location besides `memory`, `storage`, `calldata` (and return-data and code). The new data location behaves as a key-value store similar to storage with a key difference: data in transient storage is not permanent, but is scoped to the current transaction only, after which it will be reset to zero. Consequently, transient storage is as cheap as warm storage access with `TSTORE` and `TLOAD` priced at 100 gas.
```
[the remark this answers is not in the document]
matheusaaguiar: I added a small paragraph clarifying that below.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,94 @@
+---
+layout: post
+published: true
+title: 'Transient Storage Opcodes in Solidity 0.8.24'
+date: '2024-01-25'
+author: Solidity Team
+category: Explainers
+---
+
+TODO: adjust date and links!
```
cameel, line 10: This is not done yet.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,94 @@
...
+
+[Solidity 0.8.24](https://TODO/link/to/release/post) supports the opcodes included in the upcoming Cancun hardfork and, in particular, the transient storage opcodes `TSTORE` and `TLOAD` [as per EIP-1153](https://eips.ethereum.org/EIPS/eip-1153).
```
cameel, line 12:
```suggestion
[Solidity 0.8.24](https://TODO/link/to/release/post) supports the opcodes included in the upcoming Cancun hardfork and, in particular, the transient storage opcodes `TSTORE` and `TLOAD` as per [EIP-1153](https://eips.ethereum.org/EIPS/eip-1153).
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,94 @@
...
+
+Transient storage is a long-awaited feature on the EVM level that introduces another data location besides `memory`, `storage`, `calldata` (and return-data and code). The new data location behaves as a key-value store similar to storage with a key difference: data in transient storage is not permanent, but is scoped to the current transaction only, after which it will be reset to zero. Consequently, transient storage is as cheap as warm storage access with `TSTORE` and `TLOAD` priced at 100 gas.
```
cameel, line 14:
```suggestion
Transient storage is a long-awaited feature on the EVM level that introduces another data location besides `memory`, `storage`, `calldata` (and return-data and code). The new data location behaves as a key-value store similar to storage with the main difference being that data in transient storage is not permanent, but is scoped to the current transaction only, after which it will be reset to zero. Consequently, transient storage is as cheap as warm storage access, with `TSTORE` and `TLOAD` priced at 100 gas.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,94 @@
...
+
+Users should note that the compiler does not yet introduce `transient storage` as a data location which can be used as the already existing ones. However, the data location itself can be accessed using the opcodes `TSTORE` and `TLOAD` in inline assembly.
```
cameel, line 16: When we introduce it, it will likely be called just `transient`.
```suggestion
Users should note that the compiler does not yet allow using `transient` as a data location in high-level Solidity code. For the time being, data stored in this location can only be accessed using the `TSTORE` and `TLOAD` opcodes in inline assembly.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,94 @@
...
+
+An expected canonical use case for transient storage is cheaper reentrancy locks, which can be readily implemented with the opcodes as showcased below. However, given the precise specification of EIP-1153, care has to be taken for more advanced use cases of transient storage to preserve the composability of your smart contract. To raise awareness for this issue, for the time being the compiler will emit a warning on use of ``tstore`` in assembly.
```
cameel, line 18:
```suggestion
An expected canonical use case for transient storage is cheaper reentrancy locks, which can be readily implemented with the opcodes as showcased below. However, given the caveats mentioned in the specification of EIP-1153, utmost care has to be taken for more advanced use cases of transient storage to preserve the composability of your smart contract. To raise awareness for this issue, for the time being the compiler will emit a warning on use of `tstore` in assembly.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,99 @@
...
+
+We recommend to generally always clear transient storage completely at the end of a call into your smart contract to avoid these kinds of issues and to simplify the analysis of the behaviour of your contract within complex transactions. As a matter of fact, the Solidity team has been advocating for a change in the specification of transient storage to change its scope to the call frame into a smart contract within a transaction to avoid this kind of pitfall on the EVM level - however, the concern was ultimately ignored and, consequently, responsible and safe use of transient storage now lies within the responsibility of the user. We're still investigating our options to mitigate these pitfalls in future high-level language constructs built on top of the base functionality of the transient storage opcodes.
```
cameel, line 95: By the way, can you adjust the text so that there's one sentence per line? It makes it easier to make comments/suggestions for specific bits. Now I always have to include a whole paragraph.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,99 @@
...
+    function f2() nonreentrant(SECTION_A) public {
+        ...
+    }
+
+    function g1() nonreentrant(SECTION_B) public {
+        ...
+    }
+    function g2() nonreentrant(SECTION_B) public {
+        ...
+    }
```
cameel, line 89: One way to make this example less abstract would be to implement a circular buffer. Let's say, one side are donations, and the other are people picking them up. One pair of functions adds/removes from on one end, the other on the other end. Each operation requires some external call (confirmation, transfer, callback, whatever).

Should be pretty simple.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,199 @@
...
+
+contract Lender {
+    Bank public bank;
+    mapping(address => uint) public loans;
+
+    function askLoan(uint amount) public {
+        uint balance = bank.myBalance();
+        if (balance >= amount) {
+            loans[msg.sender] += amount;
+            bank.withdraw(amount);                                        // Locks the reentrancy guard and does not unlock after finishing
```
cameel, line 177: Note that this will likely be completely out of view on the blog. Code boxes are quite narrow.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,99 @@
...
+```solidity
+contract C {
+    modifier nonreentrant {
+        assembly {
+            if tload(0) { revert(0, 0) }
+            tstore(0, 1)
+        }
+        _;
+        assembly {
+            tstore(0, 0)
```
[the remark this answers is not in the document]
matheusaaguiar: Ok.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+    function f() nonreentrant public {
+        ...
+    }
```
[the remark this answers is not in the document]
matheusaaguiar: I added the usual anti-pattern in the context of a contract giving 1 `eth` gifts.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,199 @@
...
+contract Bank {
+    mapping(address => uint) public balances;
+    
+    modifier nonreentrant {
+        assembly {
+            if tload(0) { revert(0, 0) }
+            tstore(0, 1)
+        }
+        _;
+        /* Without properly resetting the lock, we cannot re-enter the contract more than one time in the same transaction
```
cameel, line 148: The reentrancy lock is an important case, but we already said above that this is a safe pattern, because the reset is pretty much an integral part of the pattern. It's unlikely that someone would do this and people will easily dismiss this example.

I mean, the example is still useful and maybe we should make it the main example in the section about reentrancy locks.

What we really need here is an example that shows how you can hurt yourself with transient storage in a use case where you'd be tempted not to reset it to save on gas. And how resetting transient storage when exiting the outermost call frame (rather than at the end of transaction) would have made it a non-issue.

Maybe a good example would be something with callbacks. Like, your contract has some function `f()` that registers something in transient storage on each call. Counts calls, tracks balance, keeps stuff passed into it on a list or something like that. Another function, `g`(), passes `f` as a callback to another contract and lets it call it a few times. Then tallies the results. You can show that if `g()` does no bother to reset transient storage then you can only call it once per transaction. Otherwise it will give you wrong results so it's no composable. If transient storage was being automatically reset on every exit from `g()`, that would not have been a problem.



 if you're not careful. And how resetting transient storage when exiting the outermost call frame (rather than at the end of transaction) would have allieviated this.

Here's one I'd suggest: a contract that has a function that's "pay per call". You call its'


 Something that will make people realize that the warning is actually warranted. IMO a good one would be some piece of code that uses transient storage to pass something down the stack without using parameters and does not bother to reset it between calls. Then show how https://github.com/ethereum/solidity-website/pull/76#discussion_r1456410706 would have avoided that problem.

For example

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: I am trying to fully understand what you suggested. Would it be something similar to this:

```solidity
contract C {
    uint constant public BSIZE = 1000;
    uint[BSIZE] circularBuffer;
    uint head = 0;
    uint tail = 0;

    modifier nonreentrant(bytes32 key) {
        assembly {
            if tload(key) { revert(0, 0) }
            tstore(key, 1)
        }
        _;
        assembly {
            tstore(key, 0)
        }
    }
    bytes32 constant SECTION_A = keccak256("key_a");
    bytes32 constant SECTION_B = keccak256("key_b");

    function AddHead() nonreentrant(SECTION_A) public payable {
        head -= 1;
        if (head == 0)
            head = BSIZE - 1;
        circularBuffer[head] = msg.value;

    }
    function RemoveHead() nonreentrant(SECTION_A) public {
        uint donation = circularBuffer[head];
        (bool success, ) = msg.sender.call{value: donation}("");
        require(success, "Failed to give back donation");
        
        head += 1
        if (head == BSIZE)
            head = 0;
        circularBuffer[head] = 0;
    }

    function AddTail() nonreentrant(SECTION_B) public {
        ...
    }
    function RemoveTail() nonreentrant(SECTION_B) public {
        ...
    }
}
```

cameel: Yeah. Will need some adjustments but the general direction is good.

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: I've come up with this very trivial (and dumb) example that can be enhanced to look more realistic:
```solidity
contract Shop {
    struct Product {
        uint product_key;
        uint quantity;
        uint price;
    }
    Product[] available_products;
    Product[] sold_products;

    function logSales(function(uint) external callback) public {
        for (uint i = 0; i < sold_products.length; ++i)
            callback(sold_products[i].quantity * sold_products[i].price);

    }
}

contract C {
    Shop public s;

    function f(uint x) external {
        assembly {
            let n := tload(0)
            tstore(0, add(n, 1))
            let total := tload(0x20)
            tstore(0x20, add(total, x)) 
        }
    }

    function g() public returns (uint mean) {
        s.logSales(this.f);
        assembly {
            let n := tload(0)
            let total := tload(0x20)
            mean := div(total, n)
        }
    }
}
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,242 @@
...
+A design is composable, if it consists of modular components that can be chained together ("composed") to more complex applications, while each component is an independent transaction that does not share state with previous components (other than global state that, to preserve composability, ought to be modified by each component atomically).
+
+For smart contracts in particular, it is important that their behaviour is self-contained in this manner, such that multiple calls into individual smart contracts can be composed to more complex applications.
+So far the EVM largely guaranteed composable behaviour, since multiple calls into a smart contract within a complex transaction are virtually indistinguishable from multiple calls to the contract stretched over several transactions (TODO: maybe mention the distinction between tx.origin and msg.sender and related bugs).
+However, transient storage allows to violate this principle and incorrect use may lead to complex bugs that only surface in use across several calls.
```
nikola-matic, line 145:
```suggestion
However, transient storage allows a violation of this principle, and incorrect use may lead to complex bugs that only surface in when used across several calls.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,242 @@
...
+author: Solidity Team
+category: Explainers
+---
+
+TODO: adjust date and links!
+
+[Solidity 0.8.24](https://TODO/link/to/release/post) supports the opcodes included in the upcoming Cancun hardfork and, in particular, the transient storage opcodes `TSTORE` and `TLOAD` as per [EIP-1153](https://eips.ethereum.org/EIPS/eip-1153).
+
+Transient storage is a long-awaited feature on the EVM level that introduces another data location besides `memory`, `storage`, `calldata` (and return-data and code).
+The new data location behaves as a key-value store similar to storage with the main difference being that data in transient storage is not permanent, but is scoped to the current transaction only, after which it will be reset to zero.
```
nikola-matic, line 15: > after which it will be reset to zero

Is this done automatically *and* always, or does it have to be done manually? Below this, we have the following statement, which is a bit contradictory:

> We recommend to generally always clear transient storage completely at the end of a call into your smart contract

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: I will add the link, almost sure it's gonna be: https://soliditylang.org/blog/2024/01/25-solidity-0.8.24-release-announcement

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+## Composability of Smart Contracts and the Danger of Transient Storage
+
+[Composability](https://en.wikipedia.org/wiki/Composability) is a basic design principle in software development in general and applies to smart contracts in particular.
+A design is composable, if it consists of modular components that can be chained together ("composed") to more complex applications, while each component is an independent transaction that does not share state with previous components (other than global state that, to preserve composability, ought to be modified by each component atomically).
+
+For smart contracts in particular, it is important that their behaviour is self-contained in this manner, such that multiple calls into individual smart contracts can be composed to more complex applications.
+So far the EVM largely guaranteed composable behaviour, since multiple calls into a smart contract within a complex transaction are virtually indistinguishable from multiple calls to the contract stretched over several transactions (TODO: maybe mention the distinction between tx.origin and msg.sender and related bugs).
```
nikola-matic, line 142:
```suggestion
So far the EVM largely guaranteed composable behaviour, since multiple calls into a smart contract within a complex transaction are virtually indistinguishable from multiple calls to the contract stretched over several transactions.
```

No need at this time I'd say.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+```
+
+No reentrant calls within the two sections are possible. I.e. an external call triggered in ``f1`` may re-enter at ``g1`` or ``g2``, but not at ``f2``.
+
+
+## Composability of Smart Contracts and the Danger of Transient Storage
+
+[Composability](https://en.wikipedia.org/wiki/Composability) is a basic design principle in software development in general and applies to smart contracts in particular. A design is composable, if it consists of modular components that can be chained together ("composed") to more complex applications, while each component is an independent transaction that does not share state with previous components (other than global state that, to preserve composability, ought to be modified by each component atomically).
```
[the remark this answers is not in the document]
r0qs: I think it is better than the wikipedia entry. So we could just replace it.

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: You can have multiple calls in the same transaction.
Transient storage is always cleared at the end of the transaction, not at the end of a call.

src/posts/2024-01-23-transient-storage.md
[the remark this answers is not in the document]
matheusaaguiar: Done.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,196 @@
...
+        }
+        circularBuffer[head] = msg.value;
+    }
+
+    function RemoveHead() nonreentrant(SECTION_A) public {
+        uint donation = circularBuffer[head];
+        (bool success, ) = msg.sender.call{value: donation}("");
+        require(success, "Failed to give back donation");
+        
+        head += 1
```
r0qs, line 106:
```suggestion
        head += 1;
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+    bytes32 constant SECTION_A = keccak256("key_a");
+    bytes32 constant SECTION_B = keccak256("key_b");
+
+    function AddHead() nonreentrant(SECTION_A) public payable {
+        head -= 1;
+        if (head == BSIZE) {
+            head = BSIZE - 1;
+        }
```
r0qs, line 97: Wait, what happens if you call `AddHead` before anything else? Wouldn't head be equals to `-1` and revert to out-of-bounds access?

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,196 @@
...
+        (bool success, ) = msg.sender.call{value: donation}("");
+        require(success, "Failed to give back donation");
+        
+        head += 1
+        if (head == BSIZE)
+            head = 0;
+        circularBuffer[head] = 0;
+    }
+
+    function AddTail() nonreentrant(SECTION_B) public {
```
r0qs, line 112: The function access `msg.value` so it should be payable.

```suggestion
    function AddTail() nonreentrant(SECTION_B) public payable {
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,196 @@
...
+
+        tail -= 1;
+        if (tail == 0)
+            tail = BSIZE - 1;
+        circularBuffer[tail] = 0;
+    }
+}
+```
+
+No reentrant calls within the two sections are possible. I.e. an external call triggered in ``f1`` may re-enter at ``g1`` or ``g2``, but not at ``f2``.
```
r0qs, line 132: I guess the example changed but the text still out-dated?

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: Yeah, that makes no sense. I was fixing that and then started to adjust other parts and patched it real quick without paying attention. The correct code here should be:
```solidity
if (head == 0) {
  head = BSIZE;
}
head -= 1;
```
But this is very crude and was just to get a feel of what @cameel was suggesting.

src/posts/2024-01-23-transient-storage.md
r0qs, line 132: Would it be this?
```suggestion
No reentrant calls within the two sections are possible. I.e. an external call triggered in ``AddHead`` may re-enter at ``AddTail`` or ``RemoveTail``, but not at ``RemoveHead``.
```

src/posts/2024-01-23-transient-storage.md
r0qs: > Yeah, that makes no sense. I was fixing that and then started to adjust other parts and patched it real quick without paying attention. The correct code here should be:
> 
> ```solidity
> if (head == 0) {
>   head = BSIZE;
> }
> head -= 1;
> ```

Hum...so maybe you can simply do:
```solidity
    function AddHead() public payable {
        head = (head + BSIZE - 1) % BSIZE;
        circularBuffer[head] = msg.value;
    }
    function RemoveHead() public {
        //...
        circularBuffer[head] = 0;
        head = (head + 1) % BSIZE;
    }
 ```
 This will make `head` goes from `BSIZE` to 0 and back, and `tail` goes from 0 to `BSIZE` and back. Not sure if this is what you really want, but it seems to be.

Comment (r0qs) [reply src/posts/2024-01-23-transient-storage.md:97]: For the tail it would be:

```solidity
    function AddTail()  public payable {
        circularBuffer[tail] = msg.value;
        tail = (tail + 1) % BSIZE;
    }
    function RemoveTail()  public {
        tail = (tail + BSIZE - 1) % BSIZE;
        uint donation = circularBuffer[tail];
        //...
    }
 ```

Comment (matheusaaguiar) [reply src/posts/2024-01-23-transient-storage.md:97]: Yes, I was trying to avoid using modular arithmetic to be as clear as possible. But that's the behaviour I think cameel suggested.

File: src/posts/2024-01-23-transient-storage.md [hunk 24]
```diff
@@ -0,0 +1,196 @@
...
+
+## Use of Transient Storage for Reentrancy Locks
+
+A reentrancy attack exploits a smart contract vulnerability in which the victim contract has its resources drained by being repeteadly entered before the balance is updated accordingly.
+In practice, what happens is that an attacker contract ``A`` deposits funds into a victim contract ``B`` and then issues a call to withdraw.
+However, the attacker contract does not implement a receive function and this causes the ``fallback`` function to be called instead.
+Inside ``fallback``, the attacker will again make a withdraw call to contract ``B``, which will cause the process to repeat itself until there are no more funds in ``B``.
+This is a know security problem and the source of a variety of bugs in smart contracts.
+In order to prevent the exploit from happening, it is recommended that all state changes, such as updating an account's balance, be made before calling external contracts and also the use of reentrancy locks/guards in the form of function modifiers. 
+The following example illustrates a simple reentrancy lock, already using the new transient storage opcodes ``TSTORE`` and ``TLOAD``:
```
nikola-matic, line 31:
```suggestion
The following example illustrates a simple reentrancy lock using the new transient storage opcodes ``TSTORE`` and ``TLOAD``:
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,196 @@
...
+
+        sentGifts[msg.sender] = false;
+        totalGifts -= 1;
+    }
+}
+```
+
+No reentrant calls to ``claimGift`` are possible.
+
+In most realistic situations, there are different code sections or functions which may be re-entered.
```
nikola-matic, line 68:
```suggestion
In more realistic situations, there are different code sections or functions which may be re-entered.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,196 @@
...
+        assembly {
+            let n := tload(0)
+            let total := tload(0x20)
+            mean := div(total, n)
+        }
+    }
+}
+```
+In this example, since function ``g`` does not clear transient storage when it ends, it cannot be called multiple times in the same transaction.
+If that were to happen, it will produce inconsistent results, because, after the first time, function ``f`` will compute results using data leftover from the previous call. If transient storage was being cleaned after each outermost call, in this case, the call to function ``g``, then this problem would not happen.
```
nikola-matic, line 196:
```suggestion
If that were to happen, it will produce inconsistent results, because, after the first time, function ``f`` will compute results using data left over from the previous call. If transient storage was being cleaned after each outermost call, in this case, the call to function ``g``, then this problem would not happen.
```

src/posts/2024-01-23-transient-storage.md
r0qs: But I'm wondering why not have only `AddHead` and `RemoveTail`? I mean, why this example must be circular? Like, based on this comment from @cameel https://github.com/ethereum/solidity-website/pull/76#discussion_r1466523345 it seems that you want to add donations in one end and pick them in the other, so it could be just a FIFO, no?

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,196 @@
...
+        s.logSales(this.f);
+        assembly {
+            let n := tload(0)
+            let total := tload(0x20)
+            mean := div(total, n)
+        }
+    }
+}
+```
+In this example, since function ``g`` does not clear transient storage when it ends, it cannot be called multiple times in the same transaction.
```
nikola-matic, line 195: I feel as though it would be helpful to add  a snippet describing what function `g` should look like in order for this to not happen (I how to clear the transient storage).

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: I think the point is that while you are operating on one side, the `head` for example, you can access the `tail`, but you cannot operate again (re-enter) on the `head` side.

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: I will add it.

src/posts/2024-01-23-transient-storage.md
cameel: Well, to be honest, I think this is way too contrived to be convincing. Still, the problem is that I have a really hard-time coming up with a good, practical example myself. I spent too much time trying to find one and I give up. I guess we have to go with this one.

I'll now stop focusing on examples and just review the rest of the post so that we can be done with it.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,222 @@
...
+The new data location behaves as a key-value store similar to storage with the main difference being that data in transient storage is not permanent, but is scoped to the current transaction only, after which it will be reset to zero.
+Consequently, transient storage is as cheap as warm storage access, with `TSTORE` and `TLOAD` priced at 100 gas.
+
+Users should note that the compiler does not yet allow using `transient` as a data location in high-level Solidity code.
+For the time being, data stored in this location can only be accessed using the `TSTORE` and `TLOAD` opcodes in inline assembly.
+
+An expected canonical use case for transient storage is cheaper reentrancy locks, which can be readily implemented with the opcodes as showcased below.
+However, given the caveats mentioned in the specification of EIP-1153, utmost care has to be taken for more advanced use cases of transient storage to preserve the composability of your smart contract.
+To raise awareness for this issue, for the time being the compiler will emit a warning on use of `tstore` in assembly.
```
cameel, line 21:
```suggestion
To raise awareness of this issue, for the time being the compiler will emit a warning on use of `tstore` in assembly.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,222 @@
...
+
+## Use of Transient Storage for Reentrancy Locks
+
+A reentrancy attack exploits a smart contract vulnerability in which the victim contract has its resources drained by being repeteadly entered before the balance is updated accordingly.
+In practice, what happens is that an attacker contract ``A`` deposits funds into a victim contract ``B`` and then issues a call to withdraw.
+However, the attacker contract does not implement a receive function and this causes the ``fallback`` function to be called instead.
+Inside ``fallback``, the attacker will again make a withdraw call to contract ``B``, which will cause the process to repeat itself until there are no more funds in ``B``.
+This is a know security problem and the source of a variety of bugs in smart contracts.
+In order to prevent the exploit from happening, it is recommended that all state changes, such as updating an account's balance, be made before calling external contracts and also the use of reentrancy locks/guards in the form of function modifiers. 
+The following example illustrates a simple reentrancy lock using the new transient storage opcodes ``TSTORE`` and ``TLOAD``:
```
cameel, line 31: I think it reads better without naming these contracts. Just call them "attacker" and "victim".

```suggestion
A reentrancy attack exploits a smart contract vulnerability in which the victim contract has its resources drained by being repeatedly entered before the balance is updated accordingly.
In practice, what happens is that the attacker contract deposits funds into the victim contract and then issues a call to withdraw.
However, the attacker contract does not implement a `receive` function and this causes his ``fallback`` function to be called instead.
Inside ``fallback``, the attacker will again make a withdrawal call to the victim contract, which will cause the process to repeat itself until there are no more funds to withdraw.
This is a known security problem and the source of a variety of bugs in smart contracts.
In order to prevent it from being exploited, it is recommended that all state changes, such as updating an account's balance, be made before calling external contracts.
Another alternative is the use of reentrancy locks/guards.

The following example illustrates a simple reentrancy lock implemented using with the help of transient storage:
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,222 @@
...
+
+Simple example:
```
cameel, line 33: This sentence seems redundant. I'd remove it.

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: I removed this and let only the test. Maybe latter we can add a better example we come up with.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,222 @@
...
+        bool sent = sentGifts[msg.sender];
+        require(!sent, "This gift was already claimed");
+        (bool success, ) = msg.sender.call{value: GIFT}("");
+        require(success, "Failed to claim Gift");
+
+        sentGifts[msg.sender] = false;
+        totalGifts -= 1;
+    }
+}
+```
```
cameel, line 64: Some simplification. Constants and messages are nice in production code but here it's simple enough that we're better off without them. Instead we should bring attention to important bits with comments.

```suggestion
```solidity
contract Generosity {
    mapping(address => bool) sentGifts;

    modifier nonreentrant {
        assembly {
            if tload(0) { revert(0, 0) }
            tstore(0, 1)
        }
        _;
        // Unlocks the guard, making the pattern composable.
        // After the function exits, it can be called again, even in the same transaction.
        assembly {
            tstore(0, 0)
        }
    }
    function claimGift() nonreentrant public {
        require(address(this).balance >= 1 ether);
        require(!sentGifts[msg.sender]);
        (bool success, ) = msg.sender.call{value: 1 ether}("");
        require(success);

        // In a reentrant function, doing this last would open up the vulnerability
        sentGifts[msg.sender] = false;
    }
}
```
```

File: src/posts/2024-01-23-transient-storage.md [hunk 32]
```diff
@@ -0,0 +1,153 @@
...
+
+No reentrant calls to ``claimGift`` are possible.
```
cameel, line 66:
```suggestion
Tnanks to the `nonreentrant` guard, no reentrant calls to `claimGift` are possible.
Such guards could already be implemented before the introduction of transient storage, using normal storage, but the high cost was discouraging.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,153 @@
...
+
+In more realistic situations, there are different code sections or functions which may be re-entered.
+In that case, it is possible that we have locks associated with these different sections and have to keep track of each individually as shown in the example below:
```
cameel, line 68:
```suggestion
Simple locks like the one above may be insufficient for complex contracts, requiring more sophisticated design patterns.
Let's consider an example, where a group of functions operates on a shared data structure (a circular buffer), while performing calls that may lead to reentrancy attempts.
Accesses to the front and end of the buffer do not interfere with each other and can be covered with separate locks, while functions accessing the same end need to share one lock to ensure atomic access.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,153 @@
...
+
+Complex example:
```
cameel: Again, this is redundant.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,147 @@
...
+        require(!sentGifts[msg.sender]);
+        (bool success, ) = msg.sender.call{value: 1 ether}("");
+        require(success);
+
+        // In a reentrant function, doing this last would open up the vulnerability
+        sentGifts[msg.sender] = false;
+    }
+}
+
+Tnanks to the `nonreentrant` guard, no reentrant calls to `claimGift` are possible.
```
aarlt, line 62: Typo: 'Tnanks' -> 'Thanks'

matheusaaguiar: Fixed. Thanks @aarlt !

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,147 @@
...
+    function RemoveTail() nonreentrant(SECTION_B) public {
+        uint donation = circularBuffer[tail];
+        (bool success, ) = msg.sender.call{value: donation}("");
+        require(success, "Failed to receive donation");
+
+        if (tail == 0)
+            tail = BSIZE;
+        tail -= 1;
+    }
+}
```
cameel, line 123: The example was fine. I simplified it a little below, but nothing in essence is changed. We can go with it. I mean, I think the original checks may have been off by one, but I did not dig into that too much. The ones below should be fine.

```suggestion
contract C {
    uint constant public BSIZE = 1000;
    uint[BSIZE] circularBuffer;
    uint head = 0; // Both always point at next empty cell.
    uint tail = 1; // Example sacrifices those two cells for simpler checks.

    modifier nonreentrant(bytes32 key) {
        assembly {
            if tload(key) { revert(0, 0) }
            tstore(key, 1)
        }
        _;
        assembly {
            tstore(key, 0)
        }
    }

    bytes32 constant HEAD_LOCK = keccak256("head");
    bytes32 constant TAIL_LOCK = keccak256("tail");

    function addHead() nonreentrant(HEAD_LOCK) public payable {
        circularBuffer[head] = msg.value;
        head = (head - 1) % BSIZE;
        require(head != tail);
    }
    function dropHead() nonreentrant(HEAD_LOCK) public {
        head = (head + 1) % BSIZE;
        require(head != tail);

        (bool success, ) = msg.sender.call{value: circularBuffer[head]}("");
        require(success);
    }

    function addTail() nonreentrant(TAIL_LOCK) public payable {
        circularBuffer[tail] = msg.value;
        tail = (tail + 1) % BSIZE;
        require(tail != head);
    }
    function dropTail() nonreentrant(TAIL_LOCK) public {
        tail = (tail - 1) % BSIZE;
        require(tail != head);

        (bool success, ) = msg.sender.call{value: circularBuffer[tail]}("");
        require(success);
    }
}
```

cameel: Ah, wait not. This still needs adjustment, because now it no longer has reentrancy problem :) We could solve that by moving the call before `head`/`tail` update, but then t would require a single lock (because `head` could overwrite `tail` and vice versa). I think it would be possible to work around that, but maybe a simpler solution is just to use two independent stacks (arrays) instead of a circular buffer...

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,147 @@
...
+```
+
+No reentrant calls within the two sections are possible. I.e. an external call triggered in ``AddHead`` may re-enter at ``AddTail`` or ``RemoveTail``, but not at ``RemoveHead``.
+Before the introduction of the new transient storage opcodes, the state of the reentrancy lock must be kept in ``storage`` which demand expensive gas costs to be read from or written to. Transient storage opcodes are cheaper and thus more suitable for the bookkepping of reentrancy guards.
+
+
+## Composability of Smart Contracts and the Danger of Transient Storage
```
aarlt, line 130: I think it would be great to have also a negative example to highlight the dangerous things. Developers may understand this much faster, if an example is shown here, instead of reading the whole text. But not sure what could be a good example.

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,147 @@
...
+This, however, can easily lead to unexpected behaviour in complex transactions, in which values set by a previous call into the contract within the same transaction remain.
+
+We recommend to generally always clear transient storage completely at the end of a call into your smart contract to avoid these kinds of issues and to simplify the analysis of the behaviour of your contract within complex transactions.
+As a matter of fact, the Solidity team has been advocating for a change in the specification of transient storage to change its scope to the call frame into a smart contract within a transaction to avoid this kind of pitfall on the EVM level - however, the concern was ultimately ignored and, consequently, responsible and safe use of transient storage now lies within the responsibility of the user.
+We're still investigating our options to mitigate these pitfalls in future high-level language constructs built on top of the base functionality of the transient storage opcodes.
+
+The use of transient storage for reentrancy locks that are cleared at the end of the call frame into the contract, are safe.
```
cameel, line 146:
```suggestion
The use of transient storage for reentrancy locks that are cleared at the end of the call frame into the contract, is safe.
```

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: Yeah, I agree. I will use two independent dynamic arrays.

cameel: Here's a second attempt, with two buffers:

```suggestion
contract C {
    uint[] bufferA;
    uint[] bufferB;

    modifier nonreentrant(bytes32 key) {
        assembly {
            if tload(key) { revert(0, 0) }
            tstore(key, 1)
        }
        _;
        assembly {
            tstore(key, 0)
        }
    }

    bytes32 constant A_LOCK = keccak256("a");
    bytes32 constant B_LOCK = keccak256("b");

    function pushA() nonreentrant(A_LOCK) public payable {
        bufferA.push(msg.value);
    }
    function popA() nonreentrant(A_LOCK) public {
        require(bufferA.length > 0);

        (bool success, ) = msg.sender.call{value: bufferA[bufferA.length - 1]}("");
        require(success);
        bufferA.pop();
    }

    function pushB() nonreentrant(B_LOCK) public payable {
        bufferB.push(msg.value);
    }
    function popB() nonreentrant(B_LOCK) public {
        require(bufferB.length > 0);

        (bool success, ) = msg.sender.call{value: bufferB[bufferB.length - 1]}("");
        require(success);
        bufferB.pop();
    }
}
```

I'll update my other suggestions around it to much this version. It's simpler and really ilustrates the reentrancy.

src/posts/2024-01-23-transient-storage.md
cameel: Variant for example with two separate buffers:
```suggestion
Simple locks like the one above may be insufficient for complex contracts, requiring more sophisticated design patterns.
Let's consider an example, where a group of functions operates on a two shared data structures, while performing calls that may lead to reentrancy attempts.
Accesses to each buffer do not interfere with the other and can be covered with separate locks, while functions accessing the same buffer need to share one lock to ensure atomic access.
```

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: Ok!

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,133 @@
...
+        require(bufferB.length > 0);
+
+        (bool success, ) = msg.sender.call{value: bufferB[bufferB.length - 1]}("");
+        require(success);
+        bufferB.pop();
+    }
+}
+```
+
+No reentrant calls within the two sections are possible. I.e. an external call triggered in ``AddHead`` may re-enter at ``AddTail`` or ``RemoveTail``, but not at ``RemoveHead``.
```
matheusaaguiar, line 112:
```suggestion
No reentrant calls within the two sections are possible. I.e. an external call triggered in ``pushA`` may re-enter at ``pushB`` or ``popB``, but not at ``popA``.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,147 @@
...
+
+        if (tail == 0)
+            tail = BSIZE;
+        tail -= 1;
+    }
+}
+```
+
+No reentrant calls within the two sections are possible. I.e. an external call triggered in ``AddHead`` may re-enter at ``AddTail`` or ``RemoveTail``, but not at ``RemoveHead``.
+Before the introduction of the new transient storage opcodes, the state of the reentrancy lock must be kept in ``storage`` which demand expensive gas costs to be read from or written to. Transient storage opcodes are cheaper and thus more suitable for the bookkepping of reentrancy guards.
```
cameel, line 113:
```suggestion
In the above we rely on transient storage being implemented as key-value storage (thus, allowing random access to any slot at the same cost) to create two separate locks, which do not interfere with each other.

No reentrant calls within the two sections are possible. I.e. an external call triggered in `popA()` may end up entering `pushB()` or `popB()` (which is perfectly safe), but not `pushB()`.
```

I'd also move the bit about cost above. I'll add it in another suggestion.

cameel: Actually, I think that the bit about cost I already added above is enough.

matheusaaguiar: Ok, I noticed that in the last sentence should be `pushA()` instead of `pushB()` in the end.

src/posts/2024-01-23-transient-storage.md
cameel: Oh, and let's give those contracts different names. I'm actually thinking about referring to this one in the next section...

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: I will fix it locally.

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: What about `DoubleBufferContract` ?

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,147 @@
...
+
+For smart contracts in particular, it is important that their behaviour is self-contained in this manner, such that multiple calls into individual smart contracts can be composed to more complex applications.
+So far the EVM largely guaranteed composable behaviour, since multiple calls into a smart contract within a complex transaction are virtually indistinguishable from multiple calls to the contract stretched over several transactions.
+However, transient storage allows a violation to this principle and incorrect use may lead to complex bugs that only surface when used across several calls.
+
+For example, since transient storage is constructed as a relatively cheap key-value store, a smart contract author may be tempted to use transient storage as a replacement for in-memory mappings without keeping track of the modified keys in the mapping and thereby without clearing the mapping at the end of the call.
```
cameel, line 126:
````suggestion
However, transient storage allows a violation to this principle and incorrect use may lead to complex bugs that only surface when used across several calls.

Let's consider a concrete example:

```solidity
contract MulService {
    function setMultiplier(uint multiplier) external {
        assembly {
            tstore(0, multiplier)
        }
    }

    function getMultiplier() private view returns (uint multiplier) {
        assembly {
            multiplier := tload(0)
        }
    }

    function multiply(uint value) external view returns (uint) {
        return value * getMultiplier();
    }
}
```

and a sequence of external calls:
```solidity
setMultiplier(42);
multiply(1);
multiply(2);
```

If the example used memory or storage to store the multiplier, it would be fully composable. It would not matter whether you split the sequence into separate transactions or grouped them in some way. You would always get the same result. This enables use cases such as batching calls from multiple transactions together to reduce gas costs. Transient storage potentially breaks such use cases since composability can no longer be taken for granted.

As another example, since transient storage is constructed as a relatively cheap key-value store, a smart contract author may be tempted to use transient storage as a replacement for in-memory mappings without keeping track of the modified keys in the mapping and thereby without clearing the mapping at the end of the call.
````

Comment (cameel) [reply src/posts/2024-01-23-transient-storage.md:126]: One more thing, I'd add this paragraph after "taken for granted":

```
Note however, that the lack of composability is not an inherent property of transient storage.
It could have been preserved if the rules for resetting its content were slightly adjusted.
Currently the clearing happens for all contracts at the same time, when the transaction ends.
If instead it was cleared for a contract as soon as no function belonging to it remained active on the call stack (which could mean multiple resets per transaction), the issue would disappear.
In the example above it would mean clearing transient storage after each of the calls.
```

File: src/posts/2024-01-23-transient-storage.md (hunk 37 above)
Comment (cameel) [reply src/posts/2024-01-23-transient-storage.md:117]: Coming up with examples has been hellish in this post :) It's really hard.

File: src/posts/2024-01-23-transient-storage.md (hunk 36 above)
Comment (cameel) [reply src/posts/2024-01-23-transient-storage.md]: Sounds good.

File: src/posts/2024-01-23-transient-storage.md (hunk 21 above)
Comment (cameel) [reply src/posts/2024-01-23-transient-storage.md]: The point was to have a reason to have two separate locks and also multiple functions sharing each lock.

But we have rewritten the example anyway.

Comment (cameel) [reply src/posts/2024-01-23-transient-storage.md]: Still, modular arithmetic was actually a good idea IMO. I actually expected that you would use it from the start :) To me it seems simpler than `if`s.

File: src/posts/2024-01-23-transient-storage.md [hunk 42]
```diff
@@ -0,0 +1,175 @@
...
+
+A reentrancy attack exploits a smart contract vulnerability in which the victim contract has its resources drained by being repeatedly entered before the balance is updated accordingly.
+In practice, what happens is that the attacker contract deposits funds into the victim contract and then issues a call to withdraw.
+However, the attacker contract does not implement a `receive` function and this causes his ``fallback`` function to be called instead.
+Inside ``fallback``, the attacker will again make a withdrawal call to the victim contract, which will cause the process to repeat itself until there are no more funds to withdraw.
+This is a known security problem and the source of a variety of bugs in smart contracts.
+In order to prevent it from being exploited, it is recommended that all state changes, such as updating an account's balance, be made before calling external contracts.
+Another alternative is the use of reentrancy locks/guards.
+
+The following example illustrates a simple reentrancy lock implemented using with the help of transient storage:
```
cameel, line 33:
```suggestion
The following example illustrates a simple reentrancy lock implemented with the help of transient storage:
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,175 @@
...
+    function claimGift() nonreentrant public {
+        require(address(this).balance >= 1 ether);
+        require(!sentGifts[msg.sender]);
+        (bool success, ) = msg.sender.call{value: 1 ether}("");
+        require(success);
+
+        // In a reentrant function, doing this last would open up the vulnerability
+        sentGifts[msg.sender] = false;
+    }
+}
```
cameel, line 60:
````suggestion
}
```
````

File: src/posts/2024-01-23-transient-storage.md [hunk 44]
```diff
@@ -0,0 +1,175 @@
+---
+layout: post
+published: true
+title: 'Transient Storage Opcodes in Solidity 0.8.24'
+date: '2024-01-25'
```
matheusaaguiar, line 5:
```suggestion
date: '2024-01-26'
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,175 @@
...
+The new data location behaves as a key-value store similar to storage with the main difference being that data in transient storage is not permanent, but is scoped to the current transaction only, after which it will be reset to zero.
+Consequently, transient storage is as cheap as warm storage access, with `TSTORE` and `TLOAD` priced at 100 gas.
+
+Users should note that the compiler does not yet allow using `transient` as a data location in high-level Solidity code.
+For the time being, data stored in this location can only be accessed using the `TSTORE` and `TLOAD` opcodes in inline assembly.
+
+An expected canonical use case for transient storage is cheaper reentrancy locks, which can be readily implemented with the opcodes as showcased below.
+However, given the caveats mentioned in the specification of EIP-1153, utmost care has to be taken for more advanced use cases of transient storage to preserve the composability of your smart contract.
+To raise awareness of this issue, for the time being the compiler will emit a warning on use of `tstore` in assembly.
```
matheusaaguiar, line 21:
```suggestion
To raise awareness of this issue, for the time being, the compiler will emit a warning on use of `tstore` in assembly.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Use of Transient Storage for Reentrancy Locks
+
+A reentrancy attack exploits a smart contract vulnerability in which the victim contract has its resources drained by being repeatedly entered before the balance is updated accordingly.
+In practice, what happens is that the attacker contract deposits funds into the victim contract and then issues a call to withdraw.
+However, the attacker contract does not implement a `receive` function and this causes his ``fallback`` function to be called instead.
+Inside ``fallback``, the attacker will again make a withdrawal call to the victim contract, which will cause the process to repeat itself until there are no more funds to withdraw.
```
matheusaaguiar, line 28:
```suggestion
Inside ``fallback``, the attacker will again make a withdraw call to the victim contract, which will cause the process to repeat itself until there are no more funds to withdraw.
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+In the above we rely on transient storage being implemented as key-value storage (thus, allowing random access to any slot at the same cost) to create two separate locks, which do not interfere with each other.
+
+No reentrant calls within the two sections are possible. I.e. an external call triggered in `popA()` may end up entering `pushB()` or `popB()` (which is perfectly safe), but not `pushA()`.
+
+
+## Composability of Smart Contracts and the Danger of Transient Storage
+
+[Composability](https://ethereum.org/developers/docs/smart-contracts/composability) is a basic design principle in software development in general and applies to smart contracts in particular.
```
cameel, line 119:
```suggestion
[Composability](https://en.wikipedia.org/wiki/Composability) is a basic design principle in software development in general and [applies to smart contracts in particular](https://ethereum.org/developers/docs/smart-contracts/composability).
```

src/posts/2024-01-23-transient-storage.md
```diff
@@ -0,0 +1,175 @@
...
+
+As another example, since transient storage is constructed as a relatively cheap key-value store, a smart contract author may be tempted to use transient storage as a replacement for in-memory mappings without keeping track of the modified keys in the mapping and thereby without clearing the mapping at the end of the call.
+This, however, can easily lead to unexpected behaviour in complex transactions, in which values set by a previous call into the contract within the same transaction remain.
+
+We recommend to generally always clear transient storage completely at the end of a call into your smart contract to avoid these kinds of issues and to simplify the analysis of the behaviour of your contract within complex transactions.
+As a matter of fact, the Solidity team has been advocating for a change in the specification of transient storage to change its scope to the call frame into a smart contract within a transaction to avoid this kind of pitfall on the EVM level - however, the concern was ultimately ignored and, consequently, responsible and safe use of transient storage now lies within the responsibility of the user.
```
cameel, line 171:
```suggestion
As a matter of fact, the Solidity team has been advocating for a change in the specification of transient storage to change its scope to the outermost call frame into a smart contract within a transaction to avoid this kind of pitfall on the EVM level - however, the concern was ultimately ignored and, consequently, responsible and safe use of transient storage now lies within the responsibility of the user.
```

cameel (approved):
ok, we're done here. There are still a few minor problems I just pointed out but I'm assuming you're patching them right now so I'm approving. The blog post is ready for tomorrow.

And thanks for great work! It turned out better than I expected in the end. It's solid.

src/posts/2024-01-23-transient-storage.md
cameel: Why not "withdrawal"? I think I suggested a change the other way just a moment ago :)

src/posts/2024-01-23-transient-storage.md
cameel: We decided we'll keep it at 25th: https://github.com/ethereum/solidity-website/pull/77#discussion_r1466955998

src/posts/2024-01-23-transient-storage.md
matheusaaguiar: Ah, you are right! Sorry, I thought it was a typo! I will fix it!

PR #90, opened 2024-03-22
Update Copyright Year in Footer

drewroberts:
I noticed the website footer had 2023 as the copyright year, so here is a simple PR to update it to 2024.

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -44,7 +44,7 @@ export const Footer: React.FC = () => (
       </Flex>
     </Flex>
     <Flex fontSize="lg" textAlign="center" flexWrap="wrap" columnGap={4}>
-      <Text>2023 Solidity Team</Text>
+      <Text>2024 Solidity Team</Text>
       <Text>•</Text>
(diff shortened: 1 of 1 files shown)


src/components/Footer.tsx
```diff
@@ -44,7 +44,7 @@ export const Footer: React.FC = () => (
       </Flex>
     </Flex>
     <Flex fontSize="lg" textAlign="center" flexWrap="wrap" columnGap={4}>
-      <Text>2023 Solidity Team</Text>
+      <Text>2024 Solidity Team</Text>
```
wackerow, line 47:
```suggestion
      <Text>{new Date().getFullYear()} Solidity Team</Text>
```

@drewroberts Thanks for this! How about doing this to make it dynamic, keeping it always up to date?

Just did the same thing over on the Remix website repo https://github.com/ethereum/remix-website/pull/70/commits/c83f94fbf91205011c844006790981a0ea320549

wackerow:
Thanks for flagging! Updated with a dynamic year so we won't need to keep doing this =). Pulling in

PR #101, opened 2024-05-13
Create 2024-05-21-solidity-0.8.26-release-announcement.md

mehtavishwa30:


src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
```diff
@@ -0,0 +1,172 @@
...
+date: '2024-05-21'
+author: Solidity Team
+category: Releases
+---
+
+We are excited to announce the release of the Solidity Compiler [v0.8.26](https://github.com/ethereum/solidity/releases/tag/v0.8.26).
+This newest version of the compiler brings support for custom errors in `require`, improved default Yul Optimizer sequence that will speed up compilation via IR, several bugfixes, and more!
+
+
+## Important Note
```
nikola-matic, line 12: What's the important note?

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Compiler Features
+
+* SMTChecker: Create balance check verification target for CHC engine.
+* Yul IR Code Generation: Cheaper code for reverting with errors of a small static encoding size.
+* Yul Optimizer: New, faster default optimizer step sequence.
+
+### Bugfixes
+
+* Commandline Interface: Fix ICE when the optimizer is disabled and an empty/blank string is used for ``--yul-optimizations` sequence.
```
cameel, line 154: Is it like that in the changelog? Do we have to correct it there as well?
```suggestion
* Commandline Interface: Fix ICE when the optimizer is disabled and an empty/blank string is used for `--yul-optimizations` sequence.
```

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
```diff
@@ -0,0 +1,172 @@
...
+
+The new sequence stops much earlier:
+![bytecode-size-vs-optimization-time-v0.8.26](https://gist.github.com/assets/137030/9c326b84-7d8e-4e5b-8413-0eaa3353b3ef)
+
+Similarly for runtime gas.
+Current sequence:
+![runtime-gas-vs-optimization-time-v0.8.25](https://gist.github.com/assets/137030/2dfdf7cf-1670-4fb9-b048-9e8f2081dc1d)
+
+New sequence:
+![runtime-gas-vs-optimization-time-v0.8.26](https://gist.github.com/assets/137030/e3a5b9c6-8441-46e2-989e-b04998d06047)
```
cameel, line 105: Images are still in my gist, not in the repo.

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
```diff
@@ -0,0 +1,172 @@
...
+            InsufficientBalance(balance[msg.sender], amount)
+        );
+        balance[msg.sender] -= amount;
+        balance[to] += amount;
+    }
+    // ...
+}
+```
+
+Note that, just like in the previously available overloads of `require`, arguments are evaluated unconditionally, so take special care to make sure that they are no expressions with unexpected side-effects.
```
cameel, line 55:
```suggestion
Note that, just like in the previously available overloads of `require`, arguments are evaluated unconditionally, so take special care to make sure that they are not expressions with unexpected side-effects.
```

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
```diff
@@ -0,0 +1,172 @@
...
+The `require` function in pre 0.8.26 versions provided two overloads:
+- `require(bool)` which will revert without any data (not even an error selector).
+- `require(bool, string)` which will revert with `Error(string)`.
+
+In this release we are introducing a new overload to support custom errors:
+- `require(bool, error)` which will revert with the custom, user supplied error provided as the second argument.
+
+Let's look at an example to understand the usage of the `require` function with custom errors:
+
+```solidity=
```
aarlt, line 30: I normally just use `solidity` - not sure weather `solidity=` is really valid..

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
cameel: By the way, note that the current plots do not render in github's preview. I do hope they will show up in the post, but it would be safer to have them in the repo even just for that reason.

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
cameel: Seems [fine in the changelog](https://github.com/ethereum/solidity/blob/v0.8.26/Changelog.md?plain=1#L14) so only need a correction here.

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
aarlt: Ok, will fix that too.

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
aarlt: I uploaded the files and updated it. Is it important to have `.svg` ending?

cameel: Probably better to have it. Any good reason not to give them proper names and extensions?

Also, a subdir in `solidity-website/public/img/` sounds like a better place for them. That's where we store the survey images as well. `src/posts/` is not where images go. Not sure Jekyll will even handle them there.

Finally, you also need absolute links I think. Check how e.g. https://github.com/ethereum/solidity-website/blob/main/src/posts/2024-04-03-solidity-developer-survey-2023-results.md handles images.

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
```diff
@@ -0,0 +1,181 @@
...
+In particular, the new sequence is susceptible to the [FullInliner Non-Expression-Split Argument Evaluation Order Bug](https://soliditylang.org/blog/2023/07/19/full-inliner-non-expression-split-argument-evaluation-order-bug/), which is not an issue for the recent versions, but would cause problems on versions older than v0.8.21.
+
+### Replacement of internal JSON Library
+
+In this release we also replaced our internal JSON library [`jsoncpp`](https://github.com/open-source-parsers/jsoncpp) with [`nlohmann::json`](https://github.com/nlohmann/json).
+
+Because of that, the formatting of the JSON output slightly changed, where it also became more strict with UTF-8 encodings.
+The old [`jsoncpp`](https://github.com/open-source-parsers/jsoncpp) allowed some invalid UTF-8 sequences, but also did not handle them properly.
+
+However, we don't expect it to create any problems because vast majority of implementation assume UTF-8 anyway.
```
cameel, line 147: How about this?
```suggestion
However, we don't expect it to create problems in practice because the vast majority of implementations assume UTF-8 anyway.
```

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
```diff
@@ -0,0 +1,181 @@
...
+The default sequence in Solidity v0.8.25 included the following steps:
+```
+dhfoDgvulfnTUtnIf [ xa[r]EscLM cCTUtTOntnfDIul Lcul Vcul [j] Tpeul xa[rul] xa[r]cL gvif CTUca[r]LSsTFOtfDnca[r]Iulc ] jmul[jul] VcTOcul jmul : fDnTOcmu
+```
+
+**Warning**: We make utmost efforts to ensure that the compiler works correctly regardless of the sequence used, employing fuzz testing to find any abnormalities, but by its very nature the default sequence is receiving a lot more coverage and problems with custom sequences are much more likely to remain undetected.
+For this reason, while the new sequence can also be used with older compilers, we recommend extreme care while doing so.
+In particular, the new sequence is susceptible to the [FullInliner Non-Expression-Split Argument Evaluation Order Bug](https://soliditylang.org/blog/2023/07/19/full-inliner-non-expression-split-argument-evaluation-order-bug/), which is not an issue for the recent versions, but would cause problems on versions older than v0.8.21.
+
+### Replacement of internal JSON Library
```
cameel, line 140:
```suggestion
### Replacement of the internal JSON library
```

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
cameel: For the file names, please just use what you see in the square brackets in the snippet above. These were actually the original file names before I uploaded them to the gist.

cameel (approved):

src/posts/2024-05-21-solidity-0.8.26-release-announcement.md
mehtavishwa30: Yeah, this is the default auto quotation in hackmd. So I missed to remove the extra from the autocorrect.

PR #112, opened 2024-09-04
Create 2024-09-04-solidity-0.8.27-release-announcement.md

mehtavishwa30:


matheusaaguiar (approved):

src/posts/2024-09-04-solidity-0.8.27-release-announcement.md
```diff
@@ -0,0 +1,132 @@
...
+### Build System
+
+* Change build system to use git submodules for some dependencies (nlohmann-json, fmtlib & range-v3).
+
+## How to Install/Upgrade?
+
+To upgrade to the latest version of the Solidity Compiler, please follow the [installation instructions](https://docs.soliditylang.org/en/v0.8.27/installing-solidity.html) available in our documentation.
+You can download the new version of Solidity here: [v0.8.27](https://github.com/ethereum/solidity/releases/tag/v0.8.27).
+
+If you want to build from the source code, do not use the source archives generated automatically by GitHub.
```
cameel, line 130: I see that this bit was simplified in this and previous release. IMO it sounds a bit weird without any explanation though. Maybe we should at least link to the right source archive?

cameel (approved):
Looks fine.

src/posts/2024-09-04-solidity-0.8.27-release-announcement.md
mehtavishwa30: I was curious too but just copied it over from the previous one that I think the team changed while I was on vacation. Should I use the original version? Not sure why it was simplified in the first place.

cameel: Well, then better to ask why it was simplified. But this is not crucial for the release so we could merge it as is and change it separately if need be.

PR #113, opened 2024-09-04
Update index.tsx

mehtavishwa30:


src/pages/index.tsx
```diff
@@ -115,12 +115,12 @@ export default function Home({
               <Box>
                 <Text lineHeight="180%" fontSize="md" mb={4}>
                   <Link
-                    href="/blog/2024/05/21/solidity-0.8.26-release-announcement/"
+                    href="https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement"
```
nikola-matic, line 118: You don't need the full URL, `/blog/2024/09/04...` will suffice.

r0qs, line 118:
```suggestion
                    href="/blog/2024/09/04/solidity-0.8.27-release-announcement/"
```
The link should be relative.

r0qs: Noticed that Nikola already replied, so resolving this.

mehtavishwa30: oh you're right. I think @r0qs suggested the same.

r0qs (approved):

PR #126, opened 2025-02-12
Add license.

ekpyron:
The repository is missing a license - I'm suggesting to add a CC-BY-SA license as in the PR.
@wackerow @mehtavishwa30 do you agree with licensing under those terms?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,349 @@
+Creative Commons Attribution-ShareAlike 4.0 International Public
+License
+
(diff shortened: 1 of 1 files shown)


mehtavishwa30 (approved):
Thanks for adding it, @ekpyron. Looks good to me, so I'm approving these changes from my side while awaiting one from @wackerow.

wackerow (approved):
Thanks @ekpyron!

PR #127, opened 2025-02-18
build: bump actions/upload-pages-artifac to v2

wackerow:
Addresses failed site deployment: https://github.com/ethereum/solidity-website/actions/runs/13384184430/job/37377833378#step:1:31

### DIFF
--- a/.github/workflows/nextjs.yml
+++ b/.github/workflows/nextjs.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Static HTML export with Next.js
         run: ${{ steps.detect-package-manager.outputs.runner }} next export -o docs
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v1
+        uses: actions/upload-pages-artifact@v2
         with:
           path: ./docs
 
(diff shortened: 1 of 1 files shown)


.github/workflows/nextjs.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - name: Static HTML export with Next.js
         run: ${{ steps.detect-package-manager.outputs.runner }} next export -o docs
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@v1
+        uses: actions/upload-pages-artifact@v2
```
r0qs, line 83: Can't we already update it to v3? https://github.com/actions/upload-pages-artifact/releases/tag/v3.0.0
Is the `upload-pages-artifact@v3` that already switched from the deprecated `actions/upload-artifact@v3` to the newer v4.


```suggestion
        uses: actions/upload-pages-artifact@v3
```

r0qs (changes requested):

.github/workflows/nextjs.yml
r0qs: @cameel we will also need to update solc-js: https://github.com/ethereum/solc-js/blob/95d3a5f953d1365388c39f60ed8d5e3ae6cabcb2/.github/workflows/scorecards.yml#L49

I will open a PR there.

r0qs:
Since we need that for the current v0.8.29 release, I'm just bumping the version and merging it ;)

PR #133, opened 2025-03-27
Create 2025-03-27-the-case-for-eof.md

mehtavishwa30:

PR #135, opened 2025-03-27
More polished wording in the EOF blog post

cameel:
Addressing comments from #133 and some left originally in https://notes.ethereum.org/@solidity/the-case-for-eof.


clonker (approved):
Nice, lgtm!

PR #139, opened 2025-04-25
docs: fix broken “using for” link

davidjsonn:
The previous link `https://ethereum.github.io/solidity/docs/using-for/` returned 404.
Replaced with the canonical Solidity docs URL:`https://docs.soliditylang.org/en/latest/contracts.html#using-for`


r0qs (approved):
Thanks @davidjsonn the action update seems unnecessary and unrelated, but it doesn't hurt to have it.

PR #166, opened 2025-12-03
ci: pin actions to commit hashes for security

davidjsonn:
https://github.com/argotorg/solidity-website/pull/139#issuecomment-3603908066


r0qs (changes requested):
Thanks a lot @davidjsonn, just a small comment.

r0qs (approved):
Thanks!

PR #176, opened 2025-12-18
Add blog post about storage array deletion bug on slot overflow

r0qs:


src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,265 @@
...
+layout: post
+published: true
+title: 'Storage Array Deletion Bug at 2^256 Boundary'
+date: '2025-12-18'
+author: Solidity Team
+category: Security Alerts
+---
+
+
+On [DATE HERE], 2024, a bug in the Solidity code generator was found by [@Audittens](https://twitter.com/Audittens). The bug was initially reported to affect deletion and partial assignment operations on storage arrays that cross the 2^256 storage boundary. During our further investigation, we discovered a third instance of the same issue affecting copying from arrays placed at the storage boundary. The bug affects both the evmasm and via-ir pipelines.
```
r0qs: I actually don't know the date when it was first reported.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,263 @@
...
+### The Bug in Legacy Codegen
+
+The legacy code generator had similar issues in `clearStorageLoop()` and `copyArrayToStorage()` functions:
+
+**Buggy Loop Condition (Assembly)**:
+```
+// Stack (right-most item being the stack top): [..., END, POS]
+// Equivalent to: POS >= END
+DUP1 DUP3 GT ISZERO
+PUSH[jumpdest]
```
cameel: `jumpdest` here looks confusingly like the `JUMPDEST` opcode.

```suggestion
PUSH[tag]
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+date: '2025-12-18'
+author: Solidity Team
+category: Security Alerts
+---
+
+
+On November 10, 2024, a bug in the Solidity code generator was found by [@Audittens](https://twitter.com/Audittens).
+The bug was initially reported to affect deletion and partial assignment operations on static (fixed-size) storage arrays that cross the 2^256 storage boundary.
+During our further investigation, we discovered a third instance of the same issue affecting copying from arrays placed at the storage boundary.
+The bug affects both the Evmasm and IR pipelines.
```
cameel:
```suggestion
The bug affects both the evmasm and IR pipelines.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+
+All three instances represent the same kind of mistake in the code generation logic: incorrect handling of storage range operations when arrays wrap around the 2^256 boundary.
+The bug is present in both compilation pipelines, as the utility functions in the IR pipeline were written with access to (and sometimes directly translated from) the ones in the Evmasm pipeline.
```
cameel:
```suggestion
The occurrences in both compilation pipelines are related, as the utility functions in the IR pipeline were written with access to (and sometimes directly translated from) the ones in the evmasm pipeline.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Background: Storage as a Circular Space
+
+Solidity's storage can be conceptualized as a ring buffer with 2^256 slots.
+Static arrays are stored contiguously. Therefore, when arrays are positioned near the storage boundary, they can wrap around from the end of the address space (i.e., Slot `2^256-1`) back to the beginning (Slot `0`).
+
+For example, an array beginning at slot `-5` (equivalent to `2^256 - 5`) with 10 `uint256` items would occupy these slots:
```
matheusaaguiar:
```suggestion
For example, an array beginning at slot `2^256 - 5` with 10 `uint256` items would occupy these slots:
```
I think it's better to just say that instead of having to say it's the same as -5.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+During our further investigation, we discovered a third instance of the same issue affecting copying from arrays placed at the storage boundary.
+The bug affects both the Evmasm and IR pipelines.
+
+All three instances represent the same kind of mistake in the code generation logic: incorrect handling of storage range operations when arrays wrap around the 2^256 boundary.
+The bug is present in both compilation pipelines, as the utility functions in the IR pipeline were written with access to (and sometimes directly translated from) the ones in the Evmasm pipeline.
+
+## Background: Storage as a Circular Space
+
+Solidity's storage can be conceptualized as a ring buffer with 2^256 slots.
+Static arrays are stored contiguously. Therefore, when arrays are positioned near the storage boundary, they can wrap around from the end of the address space (i.e., Slot `2^256-1`) back to the beginning (Slot `0`).
```
matheusaaguiar:
```suggestion
Static arrays are stored contiguously. Therefore, when arrays are positioned near the storage boundary, they can wrap around from the end of the address space (i.e., slot `2^256-1`) back to the beginning (slot `0`).
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
+---
+layout: post
+published: true
+title: 'Storage Array Deletion Bug at 2^256 Boundary'
```
cameel:
```suggestion
title: 'Lost Storage Array Write on Slot Overflow Bug'
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+
+For example, an array beginning at slot `-5` (equivalent to `2^256 - 5`) with 10 `uint256` items would occupy these slots:
+
+```
+[2^256-5, 2^256-4, 2^256-3, 2^256-2, 2^256-1, 0, 1, 2, 3, 4]
+```
+
+This wrapping behavior is well-defined in the EVM. Storage space is initially always 'clean' and gas refunds are issued when, upon halt of the program, storage that was only temporarily used is cleaned again.
+The Solidity compiler emits code that performs these cleanup operations automatically when, e.g., deleting an array or partially copying it.
+However, in the case of static arrays that wrap around the index space boundary, the generated code only performs partial cleanup, leaving residual data in storage and resulting in unintended data retention.
```
cameel: This paragraph sounds like a background on how storage wraparound works, but it also goes into what the bug was. But then not completely, because it only mentions cleanup and not copying.

I think this should focus on the circular nature of storage and show that normal EVM operations are not affected, because an index variable will overflow at `2**256` exactly the same way. On the other hand comparisons are affected, because they rely on the invariant that indices of array items grow monotonically.

Also, the cleanup refund is irrelevant to the issue or understanding of the behavior and I would not mention it at all.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+- Storage Layout Assumptions: Until the introduction of the [custom storage layout feature in Solidity 0.8.29](https://www.soliditylang.org/blog/2025/03/12/solidity-0.8.29-release-announcement), Solidity assumed that regular state variables always start at slot zero.
+- Reliance on keccak256 Hash Collision-Freeness: The compiler relies on the de-facto collision-freeness of keccak256 hashes for mappings and dynamic arrays.
+This includes the assumption that they don't overlap with slot zero, which is where contract state variables are allocated by default.
+- Undefined Behavior: Any overlap with slot zero caused by wrap-around or otherwise is therefore expected to lead to undefined and/or ill-defined behavior unless the storage layout is provided accordingly.
+
+**This bug fix addresses the clean-up inconsistency**, making corresponding operations behave consistently irrespective of whether an array wraps around the storage index space or not.
+
+## Which Contracts Are Affected?
+
+You may be affected if your contract has storage static arrays positioned such that they cross the 2^256 storage slot boundary.
```
cameel:
```suggestion
You may be affected if your contract uses an array that crosses the 2^256 storage slot boundary.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+More specifically, a contract is affected if **both** of the following conditions are met:
+
+1. The contract has a **static array** (fixed-size array) in storage positioned such that its final slot **exceeds the 2^256 storage boundary**, i.e., it wraps around.
```
cameel: It does not have to be static. It's just that getting a dynamic one in this position is extremely hard. But not impossible in theory.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+    Therefore, triggering this bug requires deliberately "tricking" the compiler through one of the following methods:
+
+    - **Inline assembly manipulation**: Explicitly setting storage pointers via assembly to place static arrays at the boundary.
+    For example, using `assembly { arr.slot := sub(0, 5) }` to position an array near the boundary.
+    - **Extremely large dynamic arrays**: Using the data area of an extremely large dynamic array or mapping that wraps around storage space.
+    This would require a dynamic array or mapping with elements that hash to positions near the boundary.
+    - **Hash collisions**: Mappings or dynamic arrays naturally hashing to boundary locations (vanishing probability).
+
+2. The contract performs one of the following operations on such arrays:
+   - **Deletion** of the array or array elements via `delete`
```
cameel: It's only deletion of an array specifically. Deletion of an element will only trigger it if that element contains an array. And it's the position of that nested array (and not the element itself) that matters.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+More specifically, a contract is affected if **both** of the following conditions are met:
+
+1. The contract has a **static array** (fixed-size array) in storage positioned such that its final slot **exceeds the 2^256 storage boundary**, i.e., it wraps around.
+In other words, the array starts at a high slot number (near `2^256 - 1`) and extends - by virtue of the ring-buffer characteristics of EVM storage - into low slot numbers (near `0`).
+
+    **Important: This bug can only be triggered by bypassing compiler safeguards.** The Solidity compiler has built-in protections that prevent static arrays from being positioned at the storage boundary under normal circumstances:
+
+    - **Natural storage layout**: Solidity's default storage allocation never produces layouts that cross the boundary.
+    - **Custom storage layout protection**: When using the custom storage layout feature (introduced in version 0.8.29), the compiler validates storage positions and raises a `TypeError: "contract extends past the end of storage when this base slot value is specified"` if you attempt to place a static array at the boundary.
```
cameel: This is really the same mechanism. It checks where the storage ends, with or without a custom layout. I'd simply say that using a layout does not affect it and is not a way to bypass this.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+**Buggy Loop Condition (Assembly)**:
+```
+// Stack (right-most item being the stack top): [..., END, POS]
+// Equivalent to: POS >= END
+DUP1 DUP3 GT ISZERO
+PUSH[tag]
+JUMPI
+```
+
+This comparison fails when `pos` starts at a high address (near 2^256) and `end` is at a low address (near zero).
```
cameel:
```suggestion
This comparison prematurely terminates the loop if `pos` starts at a high address (near `2**256`) and `end` overflows into a a low address (near zero).
The fix is conceptually the same as in the IR case.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+1. Call `fillArray()` - fills all 10 slots with values `[1, 2, 3, ..., 10]`
+2. Call `partialAssign()` - assigns `[11, 12, 13]` to first 3 elements
+
+**Buggy Behavior** (versions <= 0.8.31):
+After `partialAssign()`, calling `getArrayValues()` returns `[11, 12, 13, 4, 5, 6, 7, 8, 9, 10]` because the cleanup of the remaining 7 slots failed due to boundary crossing.
+
+**Fixed Behavior** (version 0.8.32+):
+After `partialAssign()`, calling `getArrayValues()` returns `[11, 12, 13, 0, 0, 0, 0, 0, 0, 0]` because the cleanup correctly handles the boundary wrap-around.
+
+## Affected Versions
```
cameel: This should be in the section on which contracts are affected.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+However, from a security standpoint, the **likelihood of exploitation is extremely low**.
+Running into this issue accidentally is highly unlikely because it depends on the same property that makes mappings secure: the storage address space is so enormous (2^256 slots) that it's nearly impossible to hit a specific address without deliberate manipulation.
+
+### Why Intentional Attacks Are Unlikely
+
+While an intentional attack is technically possible, several factors severely limit its usefulness to a potential attacker:
+
+1. **Difficult to hide from auditors**: The code required to exploit this bug should look suspicious to an experienced auditor, because it must be combined with other suspicious patterns:
+   - **Deliberate boundary placement**: An array that wraps around at the storage boundary must be deliberately placed at that specific location.
+   Positioning an array at slot 2^256 - N is a red flag, especially since it would typically overlap with the contract's state variables (which start at slot 0).
```
cameel:
```suggestion
       Positioning an array at slot `2**256 - N` is a clear red flag, especially since it would typically overlap with the contract's state variables (which by default start at slot 0).
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+   - **Arbitrary array access**: To make the attack less conspicuous, one could place the boundary-crossing array within a large collection of arrays.
```
cameel:
```suggestion
   - **Arbitrary array element access**: To make the attack less conspicuous, one could place the boundary-crossing array within a large collection of arrays.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+Already deployed, well-established applications are very unlikely to be vulnerable.
+The bug would need to be introduced during initial development or upgrades.
+However, a malicious owner with, for example, upgrade rights has far simpler attack options (simple backdoor function, direct fund drain, etc.).
+
+### Final Assessment
+
+Taking into account the high impact but extremely low likelihood of exploitation, we assign this bug a severity of **low** on our internal scale (very low, low, medium, high).
+This assessment reflects that while the bug is a serious correctness issue that warranted fixing, the practical security risk to deployed contracts is minimal.
+
+**Note**: Our severity assessment differs from the Ethereum Foundation Bug Bounty Program's scale (low, medium, high, critical). Our "low" assessment may map to "medium" on the program's scale.
```
cameel: I don't think this bit is relevant to our assessment. We always use our scale in our blog posts.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+        uint256[10][1] storage arr = getArray();
+        for (uint i = 0; i < 10; i++) {
+            arr[0][i] = i + 1;
+        }
+    }
+
+    function clearArray() public {
+        // BUG: In affected versions, this deletion fails to clear the array
+        delete getArray()[0];
+    }
```
cameel: You don't call this or reference it anywhere in the text.

But anyway, I think it would be better to merge this example with the previous one. IMO they're not different enough. The fact of using a layout is not that significant.

What's more important would be to show that the bug exists also in reading from a storage array. and that is not shown anywhere. You can do delete, partial assign and storage read in a single example.

cameel: Another thing that might warrant a separate example is that case we got in the submission. It's relevant, because it shows that this is technically achievable without inline assembly.

It also shows how contrived such a case is and that you have to really try hard to run into it.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+published: true
+title: 'Storage Array Deletion Bug at 2^256 Boundary'
+date: '2025-12-18'
+author: Solidity Team
+category: Security Alerts
+---
+
+
+On November 10, 2024, a bug in the Solidity code generator was found by [@Audittens](https://twitter.com/Audittens).
+The bug was initially reported to affect deletion and partial assignment operations on static (fixed-size) storage arrays that cross the 2^256 storage boundary.
```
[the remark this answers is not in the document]
r0qs: Are you suggesting the change only at this place or all occurrences of "storage boundary"? I agree that it make sense here, but for example, in this other occurrence: "... issue affecting copying from arrays placed at the storage boundary." the use seems fine to me.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: Well, yeah, hashing would likely have already caused issues, but I agree that we should be precise here.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: I believe I addressed these two comments. Please see here: https://github.com/argotorg/solidity-website/pull/176/changes/7b3b39d4cf2c482c43bd9199ee5c944edcdae1a2

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: See https://github.com/argotorg/solidity-website/pull/176/changes/7a7db66cc5aed13bbe9c1993f327158a87584f29

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,306 @@
...
+published: true
+title: 'Lost Storage Array Write on Slot Overflow Bug'
+date: '2025-12-18'
+author: Solidity Team
+category: Security Alerts
+---
+
+
+On November 10, 2024, a bug in the Solidity code generator was found by [@Audittens](https://twitter.com/Audittens).
+The bug was initially reported to affect deletion and partial assignment operations on static (fixed-size) storage arrays that cross the `2**256` slot boundary.
```
cameel: The report specifically said "fixed-length".

```suggestion
The bug was initially reported to affect deletion and partial assignment operations on fixed-length storage arrays that cross the `2**256` slot boundary.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: Changed.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: moved.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: I honestly think that it would be nice to have this information somewhere. But I agree that here may not be the best place. Maybe we should add that to our documentation: https://docs.soliditylang.org/en/v0.8.33/security-considerations.html#security-considerations

So I will be removing it as suggested.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,267 @@
...
+During our further investigation, we discovered a third instance affecting copying from arrays placed at the storage boundary.
+
+All three instances represent the same kind of mistake in the code generation logic: incorrect implementation of comparison between storage pointers that may wrap around the `2**256` boundary.
+The occurrences in both compilation pipelines are related, as the utility functions in the IR pipeline were written with access to (and sometimes directly translated from) the ones in the evmasm pipeline.
+
+We initially classified this as a minor inconsistency, since the compiler normally flags layouts approaching the storage boundary as risky.
+Combined with the extremely low likelihood of triggering it accidentally, this is why we are only addressing it now.
+Upon further review, we concluded that the behavior is incorrect regardless, and warrants fixing.
+
+We assign this bug a severity of **low** on our internal scale (very low, low, medium, high).
```
cameel:
```suggestion
We assign this bug a severity of *low* on our internal scale (very low, low, medium, high).
```

I think the article could use a little less bold text in favor of emphasis :)

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,267 @@
...
+[2**256-5, 2**256-4, 2**256-3, 2**256-2, 2**256-1, 0, 1, 2, 3, 4]
+```
+
+Since the EVM uses 256-bit integer arithmetic, this wrapping behavior is well-defined and most operations on such arrays work correctly.
+For example, reading or writing individual elements is unaffected: when computing a slot address, any overflow wraps around exactly as the storage space does.
+
+In practice, Solidity includes safeguards that make positioning an array at the storage boundary extremely unlikely:
+
+- **Static layout protection**: The compiler checks the total size of the static portion of storage layout and will not allow it to be placed at the storage boundary.
+- **Hash-based dynamic allocation**: Data areas of dynamic variables (mappings, dynamic arrays) are allocated at locations determined by hashing the position of their static part. For any variable small enough to be copied or cleared without running out of gas, it is a probabilistic impossibility for it to land anywhere near the edge of storage.
```
cameel:
```suggestion
In practice, Solidity includes safeguards that make unintentionally positioning an array at the storage boundary extremely unlikely:
- *Static layout validation*: The compiler checks the total size of the static portion of storage layout and will not allow it to extend past the storage boundary.
- *Hash-based dynamic allocation*: Data areas of dynamic variables (mappings, dynamic arrays, structs containing other dynamic variables) are allocated at locations determined by hashing the position of their static part.
    For any variable small enough to be copied or cleared without running out of gas, it is a probabilistic impossibility for it to land anywhere near the edge of storage.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,266 @@
...
+We are not aware of any deployed contracts that meet these conditions.
+
+### Affected Versions
+
+The bug has been present since early versions of the compiler:
+
+- **IR codegen**: All IR-supporting versions prior to 0.8.32
+- **evmasm codegen**: All versions prior to 0.8.32
+
+The bug was fixed in **Solidity 0.8.32** (released December 2025).
```
cameel: IMO it's this makes things sound more complicated than they are. I'd just say this:

```suggestion
The bug already existed in the very first version of the compiler (0.1.0).
All versions prior to 0.8.32 are affected.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,271 @@
...
+```yul
+function clear_storage_range(startSlot, slotCount) {
+    for { let i := 0 } lt(i, slotCount) { i := add(i, <increment>) } {
+        sstore(add(startSlot, i), 0)
+    }
+}
+```
+
+The fix is to switch from pointer-based to index-based iteration.
+This makes overflow impossible, even in the wrap-around case, because array indices can never exceed `2**256 - 1`.
```
cameel:
````suggestion
The fix is to switch from pointer-based to index-based iteration.
This makes overflow impossible, even in the wrap-around case, because array indices can never exceed `2**256 - 1`:
```yul
function clear_storage_range(startSlot, slotCount) {
    for { let i := 0 } lt(i, slotCount) { i := add(i, <increment>) } {
        sstore(add(startSlot, i), 0)
    }
}
```
````

File: src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md [hunk 25]
```diff
@@ -0,0 +1,271 @@
...
+
+### The Bug in the Evmasm Codegen
+
+The evmasm code generator had similar issues in `clearStorageLoop()` and `copyArrayToStorage()` functions:
+
+**Buggy Loop Condition (Assembly)**:
```
cameel:
```suggestion
The evmasm code generator had similar issues in `ArrayUtils::clearStorageLoop()` and `ArrayUtils::copyArrayToStorage()` functions:
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,271 @@
...
+   However, the attacker would then need to provide users access to that specific array element.
+   Contracts that allow users to access arbitrary array elements are themselves a security risk (as they enable variable overwrites) and should be flagged during audits.
+
+2. **Hash collision dependency**: Natural occurrence via hash collisions has vanishing probability.
+
+3. **Requires control over contract's implementation**: A plausible attack scenario essentially relies on the contract belonging to the attacker.
+Already deployed, well-established applications are very unlikely to be vulnerable.
+The bug would need to be introduced during initial development or upgrades.
+However, a malicious owner with, for example, upgrade rights has far simpler attack options (simple backdoor function, direct fund drain, etc.).
+
```
cameel:
```suggestion
       However, the attacker would then need to provide users access to that specific array element.
       Contracts that allow users to access arbitrary array elements are themselves a security risk (as they enable variable overwrites) and should be flagged during audits.

2. **Hash collision dependency**: Natural occurrence via hash collisions has vanishing probability.

3. **Requires control over contract's implementation**: A plausible attack scenario essentially relies on the contract belonging to the attacker.
    Already deployed, well-established applications are very unlikely to be vulnerable.
    The bug would need to be introduced during initial development or upgrades.
    However, a malicious owner with, for example, upgrade rights has far simpler attack options (simple backdoor function, direct fund drain, etc.).
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,271 @@
...
+**Buggy Behavior** (versions <= 0.8.31):
+After calling `deleteArray()`, calling `verifyCleared()` returns `false` because the deletion loop never executed or executed incorrectly, leaving some slots with non-zero values.
+
+**Fixed Behavior** (version 0.8.32+):
+After calling `deleteArray()`, calling `verifyCleared()` returns `true` because all 10 slots are correctly cleared.
+
+### Example 2: Custom Storage Layout at Boundary
+
+[Solidity's custom storage layout](https://docs.soliditylang.org/en/latest/contracts.html#custom-storage-layout) feature allows contracts to specify exact storage positions.
```
matheusaaguiar:
```suggestion
[Solidity's custom storage layout](https://docs.soliditylang.org/en/0.8.32/contracts.html#custom-storage-layout) feature allows contracts to specify arbitrary storage starting position.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,295 @@
...
+            if (arr[0][i] != 0) return false;
+        }
+        return true;
+    }
+}
+```
+
+**Reproducing the Bug**:
+1. Call `fillArray()` - fills all 10 slots with values `[1, 2, 3, ..., 10]`
+2. Call `deleteArray()` - should clear all slots
```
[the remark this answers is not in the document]
r0qs: I updated the example to fail in the buggy case and pass otherwise.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: > Another thing that might warrant a separate example is that case we got in the submission. It's relevant, because it shows that this is technically achievable without inline assembly.
> 
> It also shows how contrived such a case is and that you have to really try hard to run into it.

Nice idea. I implemented a "minimal" version of it. Needed to use 0.8.33 to compile though, because of the ICE introduced by 0.8.32. Could be avoided though by using a non-constant variable, but that would occupy a slot...now that I'm thinking more about it I could refactor it a bit haha

matheusaaguiar:
What happened to the example using custom storage layout? Was it intentionally removed?

cameel:
Yes, that's exactly what I wanted.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: haha yeah, we can change that

r0qs:
> Yes, that's exactly what I wanted.

Yeah, sorry. I deleted the comment because I realize later that you where replying to Matheus and not asking me about that example.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: Now we are not mentioning custom layouts in the blog post anymore. Not sure if we should mention it.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,286 @@
...
+- `push()` and `pop()` on dynamic arrays
+
+In practice, Solidity includes safeguards that make unintentionally positioning an array at the storage boundary extremely unlikely:
+- *Static layout validation*: The compiler checks the total size of the static portion of storage layout and will not allow it to extend past the storage boundary.
+- *Hash-based dynamic allocation*: Data areas of dynamic variables (mappings, dynamic arrays, structs containing other dynamic variables) are allocated at locations determined by hashing the position of their static part.
+    For any variable small enough to be copied or cleared without running out of gas, it is a probabilistic impossibility for it to land anywhere near the edge of storage.
+
+## Cause of the Bug
+
+The root cause is incorrect comparison logic in the generated code for storage cleanup and copy loops.
```
cameel:
```suggestion
The root cause lies in incorrect comparison logic in the generated code for storage cleanup and copy loops.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,286 @@
...
+This works correctly when array elements occupy monotonically increasing slot addresses, but fails when an array crosses the storage boundary: the start address is high (near `2**256 - 1`) while the end address is low (near `0`), causing the comparison to immediately evaluate as false and skip the loop entirely.
+
+The fix replaces pointer-based comparisons with index-based iteration, which correctly handles wraparound.
+
+## Which Contracts Are Affected?
+
+You may be affected if your contract uses an array that crosses the `2**256` slot boundary.
+More specifically, *both* of the following conditions must be met for the bug to be triggered:
+
+1. An array in storage is positioned such in a way that it crosses the `2**256` slot boundary.
```
cameel: I see I repeated the "it crosses the `2**256`" bit. Let's reword to reduce repetition and also make it more concrete:

```suggestion
1. An array in storage is positioned in such a way that its first slot is at a higher address than its last slot.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
matheusaaguiar: I think that at least should mention that it's a possible way to have arrays wrap around the storage boundary. There was a sentence like that before.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,286 @@
...
+
+The cleanup operation can be triggered:
+- Directly, using operator `delete`.
+- Through the assignment of a shorter array (*partial assignment*) - involves cleanup of the truncated area.
+- Through the removal of an element (`<array>.pop()`) - involves cleanup of that element.
+- Through the insertion of an empty element (`<array>.push()` with no arguments) - involves cleanup of that element.
+
+As explained in the Background section, Solidity does not allow explicitly declaring state variables that would overlap the storage boundary, which means that such an array has to be intentionally created in a way that bypasses the check:
+- The manipulation of the start slot of a static storage array using inline assembly.
+- Declaration of a dynamic storage array or mapping whose items are so large that the data area extends past the end of storage and contain a small static array that can be easily predicted to be placed exactly at the edge.
```
cameel:
```suggestion
- Through the manipulation of the start slot of a static storage array using inline assembly.
- Through a declaration of a dynamic storage array or mapping whose items are so large that the data area extends past the end of storage and contain a small static array that can be easily predicted to be placed exactly at the edge.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+This code works on the assumption that the array has no items at addresses lower than `startSlot`.
+When an array wraps around, this assumption is not true.
+For example when the array starts at is a high address such as `2**256 - 5` and is long enough for `endSlot` to overflow into a small number such as `5`, the condition `lt(startSlot, endSlot)` evaluates to `false` immediately, causing the loop to never execute.
+This leaves storage uncleared.
+
+The fix is to switch from pointer-based to index-based iteration.
+This makes overflow impossible, even in the wrap-around case, because array indices can never exceed `2**256 - 1`:
+```yul
+function clear_storage_range(startSlot, slotCount) {
+    for { let i := 0 } lt(i, slotCount) { i := add(i, <increment>) } {
```
cameel: This is iteration slot by slot, so the increment is always 1:
```suggestion
    for { let i := 0 } lt(i, slotCount) { i := add(i, 1) } {
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+        sstore(add(startSlot, i), 0)
+    }
+}
+```
+
+### The Bug in the Evmasm Codegen
+
+The evmasm code generator had similar issues in `ArrayUtils::clearStorageLoop()` and `ArrayUtils::copyArrayToStorage()` functions:
+```
+// Inputs on stack: END, POS <- top
```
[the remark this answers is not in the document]
cameel: Or we could do this
```suggestion
// Inputs: stack[0]=POS, stack[1]=END
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+PUSH <loop end tag>
+JUMPI
+```
+
+This comparison prematurely terminates the loop if `pos` starts at a high address (near `2**256`) and `end` overflows into a a low address (near zero).
+The fix is conceptually the same as in the IR case.
+
+## Examples
+
+### Example 1: Array Deletion Across Boundary
```
cameel:
```suggestion
### Example 1: Array Placement Using Inline Assembly
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+
+```solidity
+contract C {
+    function getArray() internal pure returns (uint256[10][1] storage arr) {
+        assembly {
+            arr.slot := sub(0, 5)  // Slot = 2**256 - 5
+        }
+    }
+
+    // Fails in buggy versions, passes in fixed versions
```
cameel: I think this comment would work better as text after the example. Here it seems just obvious and does not add much.

```
On affected compiler versions the assertions in `testDelete()` and `testPartialAssign()` do not hold.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+    function testDelete() public {
+        getArray()[0] = [uint256(1), 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        delete getArray()[0];
+        for (uint i = 0; i < 10; i++) {
+            require(getArray()[0][i] == 0, "delete failed to clear element");
+        }
+    }
+
+    // Fails in buggy versions, passes in fixed versions
+    function testPartialAssign() public {
```
cameel: We should also have a third function that triggers the bug by copying from an array.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+            require(getArray()[0][i] == 0, "partial assign failed to clear remaining elements");
+        }
+    }
+}
+```
+
+### Example 2: Triggering Without Inline Assembly
+
+The following example, based on the original bug submission, demonstrates that the bug can be triggered without inline assembly in the vulnerable contract.
+By using an extremely large array inside a mapping, one specific array element will cross the storage boundary through hash-based allocation.
```
cameel:
```suggestion
The following example, based on the original bug submission, demonstrates that the bug can be triggered even without the use of inline assembly.
By putting an extremely large array inside a mapping, we can guarantee that one specific array element will overlap the storage boundary despite the location being hash-based.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+            arr.slot := sub(0, 5)  // Slot = 2**256 - 5
+        }
+    }
+
+    // Fails in buggy versions, passes in fixed versions
+    function testDelete() public {
+        getArray()[0] = [uint256(1), 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        delete getArray()[0];
+        for (uint i = 0; i < 10; i++) {
+            require(getArray()[0][i] == 0, "delete failed to clear element");
```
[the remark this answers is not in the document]
cameel: Same in the other example.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
cameel: Same in the other example.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+    // Mirrors Vault's storage layout to calculate slot positions.
+    // This lets us find which account's balance array crosses the storage boundary.
+    mapping(string => Data) data;
+
+    event VulnerableAccountInfo(
+        uint240 account,
+        uint256 firstSlot,      // Slot of tokenIndex 0
+        uint256 lastSlot,       // Slot of tokenIndex 255
+        uint8 boundaryIndex     // Token index where the boundary is crossed
+    );
```
cameel: I'd omit the event to keep the example shorter.

Then you also don't need `lastSlot`.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+```
+
+Due to the astronomically large array (`2 ** 240` accounts × 256 token slots), exactly one account's balance array will cross the storage boundary.
+Since each account's array occupies 256 contiguous slots and the boundary is a single point (between slot `2**256-1` and slot `0`), only one 256-slot array can straddle it.
+
+The `VulnerableAccountInfo` event shows the storage layout of the vulnerable account's balance array:
+- `account`: `253663018589497636137453613405905590274105545421928947193823258725945636` (`0x24c0e260ab14f4afed6ca3d793af7ba62c7ebfa2d961fe3ea0b13d7d9124`)
+- `firstSlot`: `115792089237316195423570985008687907853269984665640564039457584007913129639871` (`0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffbf`, i.e. `2**256 - 65`)
+- `lastSlot`: `190` (`0xbe`, wrapped around)
+- `boundaryIndex`: `65` (`0x41`)
```
cameel: I'd leave only the most important bits of info here. For example I don't think showing integer values of these long numbers is very useful.
```suggestion
- `account`: `0x24c0e260ab14f4afed6ca3d793af7ba62c7ebfa2d961fe3ea0b13d7d9124`
- `boundaryIndex`: `65`
- `firstSlot`: `0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffbf`, i.e. `2**256 - boundaryIndex`
- `firstSlot + 255`: `190` (overflow)
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
r0qs: I added it for debugging while testing with Remix and thought it might be useful for anyone reading the post. But it’s unnecessary indeed, so I’ll remove it.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,294 @@
...
+
+This means token indices 0-64 occupy slots `2**256-65` through `2**256-1`, while token indices 65-255 wrap around to slots `0` through `190`.
+
+This illustrates how contrived triggering this bug is: it requires data structures large enough to wrap around the entire `2**256` storage space.
```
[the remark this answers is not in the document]
r0qs: Hahaha sometimes what looks obvious is not that obvious :P

But fair enough. I will remove it.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,289 @@
...
+For example, reading or writing individual elements is unaffected: when computing a slot address, any overflow wraps around exactly as the storage space does.
+
+In Solidity, static variables (e.g. fixed-size types and arrays, structs) are stored contiguously starting from slot 0.
+Dynamic variables (e.g. mappings, dynamic arrays) store their data at a location computed by hashing, effectively random in the vast `2**256` address space.
+For more details, see [Layout of State Variables in Storage](https://docs.soliditylang.org/en/v0.8.32/internals/layout_in_storage.html).
+
+The bug affects operations that clear or copy storage:
+- `delete` on an array
+- Array assignment, including partial assignment (e.g., assigning a 3-element array to a 10-element array clears the remaining 7 elements)
+- `push()` and `pop()` on dynamic arrays
```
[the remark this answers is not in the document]
r0qs: I restructured the background section and I think I addressed this comment.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,313 @@
...
+author: Solidity Team
+category: Security Alerts
+---
+
+
+On November 10, 2024, a bug in the Solidity code generator was found by [@Audittens](https://twitter.com/Audittens).
+The bug was initially reported to affect deletion and partial assignment operations on fixed-length storage arrays that cross the `2**256` slot boundary.
+Two instances were reported: one in the IR pipeline and one in the evmasm pipeline.
+During our further investigation, we discovered a third instance affecting copying from arrays placed at the storage boundary.
+The effect of the bug is that storage cleanup or copy operations may not be performed at all.
```
cameel:
```suggestion
The effect of the bug is that such storage cleanup or copy operations may not be performed at all.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,313 @@
...
+For example, an array beginning at slot `2**256 - 5` with 10 `uint256` items would occupy these slots:
+
+```solidity
+[2**256-5, 2**256-4, 2**256-3, 2**256-2, 2**256-1, 0, 1, 2, 3, 4]
+```
+
+Since the EVM uses 256-bit integer arithmetic, this wrapping behavior is well-defined and most operations on such arrays work correctly.
+For example, reading or writing individual elements is unaffected: when computing a slot address, any overflow wraps around exactly as the storage space does.
+
+Note that similar loops against memory and calldata are not affected, because placing an array at the end of those address spaces is virtually impossible due to gas constraints.
```
cameel: We're not talking about the bug yet so there's nothing to be affected with :)

```suggestion
Note that, while this model applies to memory and calldata as well, for all practical purposes the possibility of an overflow in those address spaces can be ignored, because placing an array at the end is virtually impossible due to gas constraints.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,313 @@
...
+### Layout of Static and Dynamic Types in Storage
+
+In Solidity, static variables (e.g. fixed-size types and arrays, structs) are stored contiguously starting from slot 0.
+Dynamic variables (e.g. mappings, dynamic arrays) store their data at a location computed by hashing, effectively random in the vast `2**256` address space.
+For more details, see [Layout of State Variables in Storage](https://docs.soliditylang.org/en/v0.8.32/internals/layout_in_storage.html).
+
+Solidity includes safeguards that make unintentionally positioning an array at the storage boundary extremely unlikely:
+- *Static layout validation*: The compiler checks the total size of the static portion of storage layout and will not allow it to extend past the storage boundary.
+- *Hash-based dynamic allocation*: Data areas of dynamic variables (mappings, dynamic arrays, structs containing other dynamic variables) are allocated at locations determined by hashing the position of their static part.
+    For any variable small enough to be copied or cleared without running out of gas, it is a probabilistic impossibility for it to land anywhere near the edge of storage.
```
cameel: If you use my previous suggestion, you can shorten this, because part of it is now redundant.
```suggestion
- *Hash-based dynamic allocation*: For any variable small enough to be copied or cleared without running out of gas, it is a probabilistic impossibility for its data area to land anywhere near the edge of storage.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,327 @@
...
+    If the target array is longer than the source, the remaining elements must be cleared.
+    For example, assigning a 3-element array to a 10-element array overwrites the first 3 elements and clears the remaining 7.
+- *Element removal*: Calling `<array>.pop()` removes the last element and clears its storage.
+    For complex types (structs, nested arrays), this triggers the cleanup loop.
+
+Copying is performed when:
+
+- *Assignment*: The source array's content is copied to the target.
+- *Element insertion*: Calling `<array>.push(value)` with an array argument copies the value to the new slot.
+    For nested arrays, this triggers the copy loop.
```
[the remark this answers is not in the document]
cameel: Actually, no. I think I was wrong about this. `push()` probably does not clear stuff, even without arguments. At least that's what the docs claim: https://docs.soliditylang.org/en/develop/types.html#array-members

I was confused, because I think it does indirectly invoke the cleanup loop when the array element is an array itself. That's due to its use of `copyArrayToStorage()`. But when that happens, the array is not being truncated so the cleanup loop will not do anything.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,337 @@
...
+
+In the compiler's implementation, cleanup and copy are separate loops.
+For example, an assignment may involve two loops: one for copying and one for cleanup.
+This means that only a fragment of an array may be affected.
```
cameel:
```suggestion
A single high-level operation may perform both copy and cleanup.
For example, a partial assignment involves first copying over the elements, then clearing unused part of the array.
When this happens for an array that crosses the storage boundary, only one of the loops will actually iterate over the boundary and wrap around.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Cause of the Bug
```
cameel:
```suggestion
## The Cause of the Bug
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,337 @@
...
+
+The root cause lies in incorrect comparison logic in the generated code for storage cleanup and copy loops.
```
cameel:
```suggestion
The root cause lies in incorrect comparison logic in the generated code for storage cleanup and some of the copy loops (the ones used for copying *from* storage).
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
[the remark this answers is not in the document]
r0qs: Yeah, it doesn’t. The issue when pushing is in the call to `storeValue`, which can invoke affected copy functions (i.e. `copyArrayToStorage` that you mentioned). When pushing with no arguments, the compiler calls `storageArrayPushZeroFunction()` in IR and `incrementDynamicArraySize()` in evmasm. Both simply increment the array length and return a reference to the new slot, if I'm not mistaken.

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Which Contracts Are Affected?
+
+You may be affected if your contract uses an array that crosses the `2**256` slot boundary.
+More specifically, *both* of the following conditions must be met for the bug to be triggered:
+
+1. An array in storage is positioned in such a way that its first slot is at a higher address than its last slot.
+1. One of the following operations is performed on the whole array or a fragment that overlaps the boundary:
+    - Cleanup (i.e., filling the area with zeros).
+    - Copying content into another storage array.
```
cameel: Oh, it should have been just "array". Only the source must be in storage.

```suggestion
    - Copying content into another array.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,337 @@
...
+
+    function testDelete() public {
+        boundaryArray()[0] = [uint256(1), 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        delete boundaryArray()[0];
+        for (uint i = 0; i < 10; i++) {
+            assert(boundaryArray()[0][i] == 0);
+        }
+    }
+
+    function testPartialAssign() public {
```
cameel:
```suggestion
    function testPartialAssignment() public {
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,337 @@
...
+        boundaryArray()[0] = [uint256(1), 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        boundaryArray()[0] = [uint256(11), 12, 13];
+        assert(boundaryArray()[0][0] == 11 && boundaryArray()[0][1] == 12 && boundaryArray()[0][2] == 13);
+        for (uint i = 3; i < 10; i++) {
+            assert(boundaryArray()[0][i] == 0);
+        }
+    }
+
+    function testCopyFrom() public {
```
cameel:
```suggestion
    function testCopyFromStorage() public {
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,322 @@
...
+        for (uint i = 0; i < 10; i++) {
+            assert(boundaryArray()[0][i] == 0);
+        }
+    }
+
+    function testPartialAssign() public {
+        boundaryArray()[0] = [uint256(1), 2, 3, 4, 5, 6, 7, 8, 9, 10];
+        boundaryArray()[0] = [uint256(11), 12, 13];
+        assert(boundaryArray()[0][0] == 11 && boundaryArray()[0][1] == 12 && boundaryArray()[0][2] == 13);
+        for (uint i = 3; i < 10; i++) {
```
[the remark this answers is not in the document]
r0qs: Haha, I mean, this is just an example of how the bug can manifest. We’re not recommending that anyone use this code in production in any way. But fine, I’ll change it :)

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,322 @@
...
+}
+```
+
+On affected compiler versions:
+- In the evmasm pipeline, the assertions in `testDelete()`, `testPartialAssign()`, and `testCopyFrom()` do not hold, while `testCopyTo()` passes.
+- In the IR pipeline, `testDelete()` and `testPartialAssign()` fail, while `testCopyFrom()` and `testCopyTo()` pass.
+
+### Example 2: Triggering Without Inline Assembly
+
+The following example, based on the original bug submission, demonstrates that the bug can be triggered even without the use of inline assembly.
```
cameel:
```suggestion
The following example, based on the original submission, demonstrates that the bug can be triggered even without the use of inline assembly.
```

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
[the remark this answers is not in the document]
r0qs: Yeah, I do it in C++. But I confess I didn’t think much about it when refactoring the examples for the third time haha. There were just too many comments to keep track of, and I was trying to finish the blog post to be able to publish it today. I was more concerned with the correctness of the examples, testing them in Remix, than with making them optimal.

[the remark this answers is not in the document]
r0qs: But you could have pushed the non-controversial changes as well. I don’t see how that would be any different from using HedgeDoc with direct edits. It would be a mess there if you only left comments suggesting changes, since there isn’t really a way to comment on specific lines. Here, it’s much clearer which lines or changes you’re suggesting, and they can be applied automatically as well. But anyway, hopefully we’re done with it now :)

src/posts/2025-12-18-lost-storage-array-write-on-slot-overflow-bug.md
```diff
@@ -0,0 +1,333 @@
...
+### Layout of Static and Dynamic Types in Storage
+
+In Solidity, the [layout of each type in storage](https://docs.soliditylang.org/en/v0.8.32/internals/layout_in_storage.html) has a static and a dynamic component to it.
+The static component is always contiguous and consists of at least one slot.
+For types that have a fixed size (value types, static arrays, structs), it stores all of the data in place.
+For dynamic arrays and mappings, its main purpose is to reserve a slot that can be used to derive unique hash-based addresses for the dynamic component, which contains the elements.
+The resulting location of the dynamic component is effectively random in the vast `2**256`-slot address space, which ensures that dynamic types can expand without overlapping other storage variables.
+
+The above scheme is applied recursively to each element of a composite type (array, struct, mapping), so a deeply nested structure can have multiple dynamic components at each nesting level.
+For example `struct S { uint x; uint[][] y; mapping (uint => uint) z}` uses 3 slots in its static component.
```
cameel: I missed a semicolon here.
```suggestion
For example `struct S { uint x; uint[][] y; mapping (uint => uint) z; }` uses 3 slots in its static component.
```

matheusaaguiar (approved):

PR #196, opened 2026-03-31
Post about the new pattern matching compiler

rodrigogribeiro:


src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,571 @@
...
+```
+Column 0 (payment)                  Action
+---------------------------------------
+Native(to, amount)           →  transfer(to, amount)
+ERC20(token, from, to, amt)  →  transferFromERC20(...)
+ERC721(token, from, to, tid) →  transferFromERC721(...)
+```
+
+Multi-scrutinee matches naturally produce multi-column matrices. This match
+function, which computes whether a state transition is permitted in an auction:
```
mbenke: The description does not match the `discount` example, perhaps it's about `canFinalize`?

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,571 @@
...
+code that was not updated after a refactor. Having the compiler detect them
+avoids the subtle class of bugs where a developer believes they are modifying
+one case of a dispatch but are actually modifying code that never executes.
+
+## The Pattern Match Compiler
+
+Exhaustiveness and redundancy checking, together with the translation of nested
+patterns into efficient runtime dispatch, are handled by a dedicated compilation
+pass in the Core Solidity prototype:
+[`DecisionTreeCompiler.hs`](https://github.com/argotorg/solcore/blob/main/src/Solcore/Desugarer/DecisionTreeCompiler.hs).
```
mbenke: The jump into compiler internals seems a bit abrupt, some introduction might be helpful.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
+---
+layout: post
+published: true
+title: "Pattern Matching in Core Solidity"
+date: "2026-xx-xx"
```
czepluch: Should be updated to fix the release date, once we've settled on one. So far I changed the name of the .md file to have the date prefixed as it's required by the framework. Would be great to get the post out around that date. (16-04-2026)

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+author: Solidity Team
+category: Announcements
+---
+
+As we described in our
+[Core Solidity deep dive](https://soliditylang.org/blog/2025/11/14/core-solidity-deep-dive/),
+Core Solidity introduces algebraic data types (ADTs) and pattern matching as
+first-class language features. This post digs into why these features matter for
+smart contract safety, and how the pattern matching compiler in the Core
+Solidity prototype is implemented.
```
czepluch: The intro starts by referencing the previous post and then saying what this post will cover. That's a soft open and the reader doesn't have a reason to care yet. Consider leading with the punch: what problem does pattern matching solve? The safety argument ("silent bugs that `require` checks can't fully prevent") is the hook, but it doesn't show up until later in the post.

I am fine to do an iteration later to address some of these "softer" details in the post.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+        uint256 amount;
+        uint256 tokenId;
+    }
+
+    function processPayment(Payment calldata payment) external {
+        if (payment.paymentType == PaymentType.NATIVE) {
+            require(payment.token == address(0), "Native: no token");
+            require(payment.amount > 0, "Native: amount required");
+            require(payment.tokenId == 0, "Native: no tokenId");
+            payable(payment.to).call{value: payment.amount}("");
```
czepluch: The return value of `.call{value: ...}("")` isn't checked here. Since the post is arguing for safety,this is a bit ironic. Consider either checking it `((bool success, ) = payable(...).call{...}(""); require(success);)` or adding a comment like `// (return value check omitted for brevity)`.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+
+```js
+data address = address(word);
+data tokenid = tokenid(word);
+
+data Payment =
+    Native(address, word)
+  | ERC20(address, address, address, word)
+  | ERC721(address, address, address, tokenid);
+```
```
czepluch, line 123: `word` appears in the data definitions but is never explained. Readers arriving at this post without having read the Core Solidity deep dive won't know what it means. A brief comment on what `word` is would be great.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+
+The pattern match binds exactly the fields that are relevant to each variant.
+There are no `require` calls to check structural invariants, the type system has
+already done that job. The `amount` field on a `Native` payment and the
+`tokenId` field on an `ERC721` payment cannot be confused, because they live in
+separate constructors.
+
+Adding a new variant, say `ERC1155(address, address, address, tokenid, word)`,
+will cause the compiler to immediately report every match expression that does
+not handle the new case. No function can silently miss the new variant.
```
czepluch, line 155: This is really nice and important. Good example to mention. An entire class of bugs eliminated :D

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+
+Adding a new variant, say `ERC1155(address, address, address, tokenid, word)`,
+will cause the compiler to immediately report every match expression that does
+not handle the new case. No function can silently miss the new variant.
+
+## Totality: Exhaustiveness as a Safety Property
+
+The example above hints at the most important safety property that pattern
+matching brings to smart contracts: **exhaustiveness checking**, also called
+totality. A pattern match is exhaustive if every possible value of the scrutinee
```
czepluch: "Scrutinee" is used several times in the post but never defined. Most Solidity devs won't know this term. Consider defining it on first use here, e.g.: "...every possible value of the scrutinee (the expression being matched on) is handled by at least one branch." 
Alternatively, using a more layman's explanation like "every possible value of the matched expression is handled by at least one branch"

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+
+The example above hints at the most important safety property that pattern
+matching brings to smart contracts: **exhaustiveness checking**, also called
+totality. A pattern match is exhaustive if every possible value of the scrutinee
+type is handled by at least one branch. The Core Solidity compiler enforces this
+statically and rejects any program that contains an incomplete match.
+
+Why does this matter so much for smart contracts in particular? Smart contract
+bugs are extremely costly to fix. Once a contract is deployed, its logic is
+immutable (absent an upgradeable proxy pattern), and any ether or tokens locked
```
czepluch: nit: "(absent an upgradeable proxy pattern)" reads a bit formal. Consider "(unless using an upgradeable proxy)", which might be more natural for the audience.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+statically and rejects any program that contains an incomplete match.
+
+Why does this matter so much for smart contracts in particular? Smart contract
+bugs are extremely costly to fix. Once a contract is deployed, its logic is
+immutable (absent an upgradeable proxy pattern), and any ether or tokens locked
+by a buggy contract may be permanently inaccessible. In the Classic Solidity
+`calculateFee` function above, the final `return 0` is dead code that exists
+only because the compiler cannot verify that the if-else chain covers all
+variants. If a future developer adds a fourth `PaymentType` and forgets to
+update `calculateFee`, the function silently returns zero, a potentially
```
czepluch:
```suggestion
update `calculateFee`, the function silently returns zero. This is a potentially
```
this is a comma splice (two clauses joined by just a comma)

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+avoids the subtle class of bugs where a developer believes they are modifying
+one case of a dispatch but are actually modifying code that never executes.
+
+## The Pattern Match Compiler
+
+Exhaustiveness and redundancy checking, together with the translation of nested
+patterns into efficient code, are handled by a dedicated compilation pass in the
+Core Solidity prototype:
+[`DecisionTreeCompiler.hs`](https://github.com/argotorg/solcore/blob/main/src/Solcore/Desugarer/DecisionTreeCompiler.hs).
+The implementation follows the ideas described in paper
```
czepluch:
```suggestion
The implementation follows the ideas described in the paper
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
czepluch: i agree. The first part of the blog post up until this point is more explanatory of "what pattern matching and ADT is" and from here on it's more focused on how it works. I am not sure if the audience for those two things are the same. 
We might not have to make to separate blogposts, but we should definitely mention here that the remainder of the post is diving deeper into how we've implemented it and isn't necessarily relevant if the reader only wanted to understand what it is.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+and prune the matrix faster.
+
+For the `discount` matrix above:
+
+```
+Column 0 (state) necessity: 2  (Active, Active — two non-variable patterns)
+Column 1 (phase) necessity: 2  (Early, Late — two non-variable patterns)
+```
+
+Both columns score equally. When scores tie the compiler prefers the shallowest
```
czepluch:
```suggestion
Both columns score equally. When scores tie, the compiler prefers the shallowest
```

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+Column 1 (phase) necessity: 2  (Early, Late — two non-variable patterns)
+```
+
+Both columns score equally. When scores tie the compiler prefers the shallowest
+scrutinee, the one that requires the fewest field accesses to reach from the
+top-level value, so column 0 is selected first.
+
+For a matrix that is more clearly asymmetric, consider:
+
+```js
```
czepluch, line 290: question: Is it on purpose that the code inside the `match {}` block is not indented?

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+Column 0 (state)    Column 1 (approved)   Action
+-------------------------------------------------
+Ended(_, _)         True           →  return True
+Cancelled(_, _)     True           →  return True
+$v0                 $v1            →  return False
+```
+
+Here column 0 has necessity score 2, column 1 has score 1, so column 0 is
+selected and the compiler emits a single `switch` on the auction state tag
+before ever looking at `approved`.
```
czepluch: I don't understand how you get to column 1 having a score of 1. Can you explain how we arrive at that? Isn't (True, True, $v1) two non-variable patterns? 
Maybe I'm just not used enough to looking at this, but either way, being more explicit about how we arrive at the result would help readers understand

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+
```
czepluch: the `canFinalize` example seems incomplete compared to the `discount` example, which had an explanation of the results.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+the flattening (`$bid`, `$pad`) has been substituted into the arm body. The
+compiler carries an _occurrence map_ that tracks which sub-expression each
+variable is bound to, and applies a single substitution pass over every arm body
+to install those bindings before emitting code.
+
+### Lowering to Yul
+
+After pattern compilation, the program goes through specialization
+(monomorphization of generic functions and type class instances) and is emitted
+as Hull: a first-order functional intermediate representation with sum and
```
czepluch, line 363: why is it called Hull? Can I read more about it anywhere? Readers might have questions.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+1 (tag) + max(size of each constructor's payload)
+```
+
+The tag is a single word that identifies which constructor is active (starting
+at 0). The payload slots hold the fields of the active constructor, padded with
+unused words to fill the maximum payload size across all constructors.
+
+**Binary sums (`A | B`).** A type with two constructors is represented as a
+single word: 0 (`false` in Yul) for the first constructor, 1 (`true` in Yul) for
+the second. Because both `unit` payload fields are zero-sized, no payload slot
```
czepluch: Only mention of `unit`. Consider adding an explanation of what it is. Currently it sounds like something everyone knows what is, but I'm not sure that's the case.

czepluch: Maybe even better yet is to just remove it, since it's only mentioned this once. Could a phrase like "Because
  neither False nor True carries any data, no payload slot is needed beyond the tag." be used instead?

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+is needed beyond the tag. So `Bool = False
+| True` occupies exactly one stack
```
czepluch: Maybe not a problem, but having the inline code span two lines like this looks a bit weird and I don't know if all markdown renderers support it

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+
+```js
+data AuctionState =
+    NotStarted(uint256)
+  | Active(uint256, address)
+  | Ended(uint256, address)
+  | Cancelled(uint256, address);
+```
+
+is initially encoded as nested binary sums: `NotStarted` becomes `inl`, and the
```
czepluch: I have no idea if `inl` is something I'm supposed to know what is or not. It came out of nowhere. 
Consider either briefly defining them ("inject left"/"inject right") or just describing the concept without the jargon: "NotStarted is tagged as the first alternative, and the remaining constructors are nested inside the second alternative."

czepluch (changes requested):
A bit of general feedback:

I think the post is good as a whole, good and detailed explanations and examples.

Terminology consistency: "match expression," "match statement," and just "match" are used interchangeably across the post. Worth picking one and sticking with it.

Post structure / audience split: The post is effectively two documents, a user-facing feature announcement (up through "Totality:
  Exhaustiveness as a Safety Property") and a compiler engineering deep dive ("The Pattern Match Compiler" onward). Most Solidity devs will drop off at that boundary. We should consider either splitting into two posts, or adding a note at the start of "The Pattern Match Compiler" that readers can skip to the conclusion if they're primarily interested in using the feature rather than how it's implemented.

Missing gas discussion: The conclusion mentions "no extra overhead" but gas cost is never addressed directly. For Solidity devs, gas is always the first question. Even one sentence in the "Lowering to Yul" section or the conclusion confirming that pattern matching compiles to the same switch you'd write by hand would help.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+[Compiling Pattern Matching to Good Decision Trees](http://moscova.inria.fr/~maranget/papers/ml05e-maranget.pdf).
+
+This pass runs after type inference and before specialization. Its job is to
+transform `match` expressions over arbitrary nested patterns into a _decision
+tree_, a form where each node tests exactly one scrutinee against flat
+constructor patterns, with no nesting. The resulting tree is then converted back
+into a simplified `match` statement that later passes (specialization and Yul
+emission) can handle directly.
+
+### The Pattern Matrix
```
mbenke: Before introducing the matrix, it might be worthwhile to mention why (or rather when) this is a nontrivial problem: matching on a single variable is easy, matching on multiple is not. 

If we split the post in two as Jakob suggests, it might be a good teaser for the second post: show a nontrivial (but perhaps not too big) example and challenge readers to translate it to nested simple matches, then discuss the solutions in the second post.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+[`DecisionTreeCompiler.hs`](https://github.com/argotorg/solcore/blob/main/src/Solcore/Desugarer/DecisionTreeCompiler.hs).
+The implementation follows the ideas described in paper
+[Compiling Pattern Matching to Good Decision Trees](http://moscova.inria.fr/~maranget/papers/ml05e-maranget.pdf).
+
+This pass runs after type inference and before specialization. Its job is to
+transform `match` expressions over arbitrary nested patterns into a _decision
+tree_, a form where each node tests exactly one scrutinee against flat
+constructor patterns, with no nesting. The resulting tree is then converted back
+into a simplified `match` statement that later passes (specialization and Yul
+emission) can handle directly.
```
mbenke: Might be worth mentioning *what* can be handled by Yul: simple numeric switches of the form

```
switch calldataload(4)
    case 0 {
        x := calldataload(0x24)
    }
    default {
        x := calldataload(0x44)
    }
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,777 @@
...
+Native(to, amount)           →  transfer(to, amount)
+ERC20(token, from, to, amt)  →  transferFromERC20(...)
+ERC721(token, from, to, tid) →  transferFromERC721(...)
+```
+
+Multi-scrutinee matches naturally produce multi-column matrices. This match
+function, which computes whether a state transition is permitted in an auction:
+
+```js
+data Phase = Early | Late
```
mbenke:
```suggestion
data Phase = Early | Late ;
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Fixed.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: Fixed.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Removed.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: Added a sentence saying that an explanation about the intermediate languages used in core solidity compilation will be the subject of future posts.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Removed.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Fixed.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Removed.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+category: Announcements
+---
+
+As we described in our
+[Core Solidity deep dive](https://soliditylang.org/blog/2025/11/14/core-solidity-deep-dive/),
+Core Solidity introduces algebraic data types (ADTs) and pattern matching as
+first-class language features. This post explains why these features matter for
+smart contract safety — not just as language niceties, but as a way to eliminate
+an entire class of bugs at compile time that today costs real money to find and
+fix.
```
czepluch: It would be better to lead with something a bit more "catching" rather than referencing a previous blog post. Maybe something like:

>  Smart contracts often need to handle values that come in several variants: a payment might be native ETH, an ERC20 transfer, or an NFT transfer; an auction can be not-started, active, ended, or cancelled. Each variant needs different handling, and the logic for that handling tends to be scattered across many functions. When a developer adds a new variant and forgets to update one of those functions, the compiler says nothing. Once deployed, that oversight can cost real money to find and fix.
>
> Core Solidity's algebraic data types (ADTs) and pattern matching, introduced in our [deep dive post](https://soliditylang.org/blog/2025/11/14/core-solidity-deep-dive/), eliminate this class of bug at compile time. This post explains why that matters for contract safety and how the feature holds up in practice.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+constraint that, e.g., a `NATIVE` payment has no `token`, each function must
+repeat those structural checks. This is both verbose and error-prone: a
+developer might add a new function and forget to validate one of the fields.
+
+## Algebraic Data Types Make Invalid States Unrepresentable
+
+Core Solidity lets us define a `Payment` type that expresses these three
+variants precisely. (`word` is Core Solidity's primitive type for a raw 256-bit
+EVM word, the same type for a Yul variable.)
+
```
czepluch, line 114: > `word` is Core Solidity's primitive type for a raw 256-bit EVM word, the same type for a Yul variable.

This reads a bit weird. I suggest rephrasing to something a bit simpler: "`word` is Core Solidity's primitive 256-bit type, equivalent to an untyped Yul variable."

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+```
+
+The pattern match binds exactly the fields that are relevant to each variant.
+There are no `require` calls to check structural invariants, the type system has
+already done that job. The `amount` field on a `Native` payment and the
+`tokenId` field on an `ERC721` payment cannot be confused, because they live in
+separate constructors.
+
+Adding a new variant, say `ERC1155(address, address, address, tokenid, word)`,
+will cause the compiler to immediately report every match expression that does
```
czepluch: Here we use the term `match expression`, but on line [246](https://github.com/argotorg/solidity-website/pull/196/changes#diff-c5f8b77050087a8a123d7c626f5c2a62f64577358bf06f6dcaf866a8052cb87fR246) we use `match statement`. Let's be consistent. Or is there a subtle difference I'm not aware of?

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+not handle the new case. No function can silently miss the new variant.
+
+## Totality: Exhaustiveness as a Safety Property
+
+The example above hints at the most important safety property that pattern
+matching brings to smart contracts: **exhaustiveness checking**, also called
+totality. A pattern match is exhaustive if every possible value of the scrutinee
+(the value being matched on) type is handled by at least one branch. The Core
```
czepluch:
```suggestion
type (the value being matched on) is handled by at least one branch. The Core
```

Reads better like this. Letting "scrutinee type" stand alone and not be split in the middle.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+is not enforced by the compiler, so it can fail. When a team member adds
+`ERC1155` support under deadline pressure and misses one function, the compiler
+says nothing. Pattern matching moves that discipline into the toolchain, where
+it cannot be forgotten.
+
+## The Pattern Match Compiler
+
+Exhaustiveness and redundancy checking, together with the translation of nested
+patterns into efficient code, are handled by a dedicated compilation pass in the
+Core Solidity prototype and it follows the ideas described in
```
czepluch:
```suggestion
patterns into efficient code, are handled by a dedicated compilation pass in the
Core Solidity prototype, which follows the ideas described in
```

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+Exhaustiveness and redundancy checking, together with the translation of nested
+patterns into efficient code, are handled by a dedicated compilation pass in the
+Core Solidity prototype and it follows the ideas described in
+[Compiling Pattern Matching to Good Decision Trees](http://moscova.inria.fr/~maranget/papers/ml05e-maranget.pdf).
+
+This pass runs after type inference and before code generation. Its job is to
+transform `match` expressions over arbitrary nested patterns into a _decision
+tree_ — a form where each node tests exactly one scrutinee against flat
+constructor patterns, with no nesting. The resulting tree is then converted back
+into simplified `match` statements that the Yul backend can handle directly.
```
czepluch: See comment on line 141 about consistent terminology

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+Warning: Clause (True → return True) is redundant.
+  in function f
+  in match (x)
+```
+
+Unlike non-exhaustive matches, redundant-clause warnings do not prevent
+compilation: they are surfaced as warnings to assist developers during
+refactoring.
+
+## Lowering to Yul — and Why There Is No Overhead
```
czepluch:
```suggestion
## Lowering to Yul (and Why There Is No Overhead)
```

This one is very much a nit. Em-dashes are looked upon weirdly in this AI age and most people think it's a tell that AI wrote things. I do like em-dashes though, so I'm fine with leaving the title as is.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+    case 1 { _result := false; leave }   // Active
+    case 2 { _result := true;  leave }   // Ended
+    case 3 { _result := true;  leave }   // Cancelled
+}
+```
+
+This is the same code you would write by hand if you were implementing a tagged
+union in Yul directly. The new language / compiler machinery: the ADT
+definition, the exhaustiveness check, the pattern matrix compilation contributes
+adds zero instructions to the final bytecode.
```
czepluch:
```suggestion
union in Yul directly. This machinery, the ADT definition,
the exhaustiveness check, the pattern matrix compilation,
adds zero instructions to the final bytecode.
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+Core Solidity's algebraic data types and pattern matching address a class of
+smart contract bugs that Classic Solidity has no good defense against: the
+silent handling of missing or newly-added cases.
+
+The problems are structural. When a contract's logic depends on a set of
+alternatives - payment types, auction phases, token standards, vote outcomes -
+Classic Solidity provides enums and structs, but it cannot enforce that every
+function that dispatches on that type handles all cases, or that every variant
+carries the right fields and no others. Developers fill the gap with runtime
+`require` checks and a discipline of adding cases everywhere. That discipline is
```
czepluch: nit: it seems that this section is using hyphens (`-`) instead of em-dashes (`—`).
It's a small detail, but it caught my eye when looking at rendered blogpost.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
czepluch:
```suggestion
first-class language features. This post explains why these features matter for
smart contract safety, and how they eliminate
an entire class of bugs at compile time that today costs real money to find and
fix.
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+            return payment.amount / 10;
+        } else if (payment.paymentType == PaymentType.ERC721) {
+            return 0.01 ether;
+        }
+        return 0; // unreachable, but the compiler requires it
+    }
+}
+```
+
+This code has several safety problems that the language cannot help us with.
```
czepluch:
```suggestion
This code has several safety problems that the language cannot catch.
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+represent any of the three payment variants, but nothing in the type prevents us
+from filling those fields with incoherent values. The `require` calls in
+`processPayment` are our only defense, and they have to be repeated in every
+function that receives a `Payment`.
+
+**No compile-time totality guarantee.** If we add a new variant, say `ERC1155`,
+to the `PaymentType` enum, the compiler will not tell us where the dispatch
+logic needs to be updated. `processPayment` will silently do nothing for
+`ERC1155` payments. `calculateFee` will silently return 0. Every function that
+dispatches on `PaymentType` is a potential silent failure waiting to happen.
```
czepluch:
```suggestion
dispatches on `PaymentType` can silently fail.
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+## Totality: Exhaustiveness as a Safety Property
+
+The example above hints at the most important safety property that pattern
+matching brings to smart contracts: **exhaustiveness checking**, also called
+totality. A pattern match is exhaustive if every possible value of the scrutinee
+(the value being matched on) type is handled by at least one branch. The Core
+Solidity compiler enforces this statically and rejects any program that contains
+an incomplete match.
+
+Why does this matter so much for smart contracts in particular? Smart contract
```
czepluch:
```suggestion
Smart contract
```

I'd remove the rhetorical question. Doesn't add much value.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+dispatches on a type covers all cases — manually tracing if-else chains and
+checking that no variant falls through to a silent default. This is tedious,
+error-prone work that scales with the number of functions and variants in a
+codebase.
+
+Pattern matching with exhaustiveness checking makes that entire category of
+audit finding disappear. When the compiler rejects incomplete matches, an
+auditor does not need to check whether `calculateFee` handles `ERC1155` — if it
+compiled, it does. The time auditors previously spent tracing dispatch logic can
+be spent on higher-value findings. For projects paying $50,000–$500,000 for an
```
czepluch:
```suggestion
be spent on higher-value findings. For projects paying five or six figures for an
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+audit, that is a concrete, measurable saving.
```
czepluch:
```suggestion
audit, that is a real saving.
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+
+### What About Existing Validation Patterns?
+
+A reasonable question from developers who already write defensive code: "I
+already use `require` checks and careful enum handling. What does this buy me?"
+
+The answer is not that your current practices are wrong — it is that they do not
+scale. Today, the discipline of "update every dispatch function when you add a
```
czepluch:
```suggestion
Your existing practices aren't wrong, they just don't scale.
Today, the discipline of "update every dispatch function when you add a
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+variant" lives in your head, in code review checklists, and in audit reports. It
```
czepluch:
```suggestion
variant" relies on memory, code review checklists, and audit reports. It
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,517 @@
...
+
+Exhaustiveness and redundancy checking, together with the translation of nested
+patterns into efficient code, are handled by a dedicated compilation pass in the
+Core Solidity prototype and it follows the ideas described in
+[Compiling Pattern Matching to Good Decision Trees](http://moscova.inria.fr/~maranget/papers/ml05e-maranget.pdf).
+
+This pass runs after type inference and before code generation. Its job is to
+transform `match` expressions over arbitrary nested patterns into a _decision
+tree_ — a form where each node tests exactly one scrutinee against flat
+constructor patterns, with no nesting. The resulting tree is then converted back
```
czepluch:
```suggestion
tree_, a form where each node tests exactly one scrutinee against flat
constructor patterns. The resulting tree is then converted back
```

nit: "with no nesting" seems redundant after you already say it's "flat"

czepluch (changes requested):
The post is much better overall now. 

My review mainly addresses formulation and grammar fixes.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Fixed.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: It is wrong. The correct is match statement. Fixed.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: Fixed.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+
+This code has several safety problems that the language cannot help us with.
+
+**Invalid states are representable.** The `Payment` struct can be constructed
+with contradictory fields: a `NATIVE` payment with a non-zero `token` address,
+or an `ERC721` payment with no `tokenId`. The struct has enough fields to
+represent any of the three payment variants, but nothing in the type prevents us
+from filling those fields with incoherent values. The `require` calls in
+`processPayment` are our only defense, and they have to be repeated in every
+function that receives a `Payment`.
```
cameel: I'd highlight also that they cost gas, unlike type-system-level checks. The compiler description goes in depth about the low overhead of the abstraction but I don't see it mentioned in the part about benefits.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+        uint256 amount;
+        uint256 tokenId;
+    }
+
+    function processPayment(Payment calldata payment) external {
+        if (payment.paymentType == PaymentType.NATIVE) {
+            require(payment.token == address(0), "Native: no token");
+            require(payment.amount > 0, "Native: amount required");
+            require(payment.tokenId == 0, "Native: no tokenId");
+            (bool success, ) = payable(payment.to).call{value: payment.amount}("");
```
cameel: It would make more sense to declare `to` as `address payable` in the struct since it's an address that's meant to receive payments.

cameel: Same for Core Solidity version actually. Does it have `address payable`?

And what's `transfer()` that you use in it? A stdlib function or something user-defined? I guess there's no syntax sugar for `<address>.call()` yet?

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+bugs are extremely costly to fix. Once a contract is deployed, its logic is
+immutable (absent an upgradeable proxy pattern), and any ether or tokens locked
+by a buggy contract may be permanently inaccessible. In the Classic Solidity
+`calculateFee` function above, the final `return 0` is dead code that exists
+only because the compiler cannot verify that the if-else chain covers all
+variants. If a future developer adds a fourth `PaymentType` and forgets to
+update `calculateFee`, the function silently returns zero — a potentially
+expensive mistake that will not be caught until it is too late.
+
+Exhaustiveness checking transforms this category of bug into a compile-time
```
cameel: Is `category of bug`/`class of bug` grammatical? It sounds less natural to me than `category of bugs`/`class of bugs`, though I'm not certain if it's because it's incorrect or just uncommon.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+    | NotStarted(reserve) =>
+        require(msg.value >= reserve);
+        return Active(msg.value, msg.sender);
+    | Active(currentBid, bidder) =>
+        require(msg.value > currentBid);
+        transferFunds(bidder, currentBid);
+        return Active(msg.value, msg.sender);
+    }
+}
+```
```
cameel, line 195: This example is fine, but I just wanted to mention that another (perhaps more idiomatic) option here would be error handling. This is the classic situation where forgetting an important case is common and can have disastrous consequences.

It would be something along these lines:
```haskell
data Result =
    Ok(T)
  | Error(uint256);

function doStuff() -> T {
    match stuff() {
    | Ok(result) => return result;
    }
}
```

Though the issue I see after writing it down is that you'd be unlikely to use `match` here unless you already remember that you have to handle the error. Core Solidity has no equivalent for Rust's `#[must_use]` or C++'s `[[nodiscard]]`, does it? :)

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+
+The algorithm works by treating the match arms as a _pattern matrix_ (one row
+per arm, one column per scrutinee), selecting the most informative column to
+test first using a necessity heuristic, and recursively building sub-matrices
+for each constructor case. Exhaustiveness and redundancy are both detected as
+natural byproducts of this construction: a missing case surfaces when the matrix
+has no row to cover a particular input, and a redundant arm surfaces when its
+row is already subsumed by earlier rows.
+
+The practical output of this pass is straightforward. The `discount` function:
```
cameel: This sounds as if I should already be familiar with the `discount` function. Would be better to introduce it as an example. E.g.:

```suggestion
The practical output of this pass is straightforward.
For example, let's take a look at the following function:
```

cameel: Or did you actually mean to reference something from a previous example? This function does look somewhat similar to `calculateFee()`.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+
+```js
+data Phase = Early | Late;
+
+function discount(state : AuctionState, phase : Phase) -> uint256 {
+    match state, phase {
+    | Active(bid, _), Early => return bid / 10;
+    | Active(bid, _), Late  => return bid / 20;
+    | _, _                  => return 0;
+    }
```
cameel, line 284: I think that before jumping into the compiler description, we could use a section explaining the `match` syntax and capabilities. For example I had no idea that it supported multiple values, only learning it from this example. The `_` may be self-explanatory, but still would not hurt to explicitly tell people what it is. And `$` comes out of the blue. I assume this is just to show it's pseudo-code?

It would also make the transition from the high-level benefits of `match` to implementation details less abrupt.

cameel: You could use it also to introduce some Core Solidity-specific terms. For example I'm not sure it will be clear to everyone what "constructor"/"constructor pattern" exactly means here. It may eventually become clear from context, but it will probably be confusing at first, because it has a completely different meaning in Classic Solidity.

Similar issue with "newtype" used later in the text.

I mean, I don't think it will confuse advanced users who know more about languages, but one of the complaints I've seen repeated after previous posts was that they were too abstract. So for a wider reach I think it's best to assume that the reader only knows Classic Solidity.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+runtime cost. Pattern matching over algebraic data types does not. The generated
+Yul is the same `switch` statement a careful developer would write by hand.
+
+After pattern compilation, the program goes through specialization
+(monomorphization of generic functions and type class instances) and is emitted
+as Hull: a first-order functional intermediate representation with sum and
+product types. The separate `yule` backend, implemented in
+[`Translate.hs`](https://github.com/argotorg/solcore/blob/main/yule/Translate.hs),
+then translates Hull into Yul. The intermediate languages used throughout the
+Core Solidity compilation pipeline will be the subject of future posts.
```
cameel, line 367: This paragraph is extremely dense and I'm not sure it will be understood by our intended audience. I mean, all the necessary information is there, it's just how it's written. IMO at least this information should be conveyed in simpler terms:
- Hull is an intermediate language
- Hull is a step between SAIL and Yul where we no longer have generics
- `yule` is a component of the Core Solidity compiler
- `Translate.hs` is just the file where you find the implementation (TBH at first it sounds like some library, like `ethers.js` or something)

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+
+### Runtime Representation of Sum Types
+
+Yul has no algebraic types, only 256-bit words. Every sum type is therefore
+flattened onto the EVM stack as:
+
+```
+1 (tag word) + max(payload size across all constructors)
```
cameel, line 375: This is meant to show the stack, right? I think it would be clearer if it was written vertically. Maybe like this:
```
|--------------|
| tag word     | 0
|--------------|
| payload slot | 1
|--------------|
| payload slot | 2
|--------------|
|   ...        | ...
|--------------|
| payload slot | max(payload size across all constructors)
|--------------|
```

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+    }
+}
+```
+
+```yul
+function usr$not(_v0) -> _result {
+  switch _v0
+    case false { _result := true;  leave }
+    case true  { _result := false; leave }
+}
```
cameel, line 425: Let's stick to 4-space indents. That's our convention in Yul and TBH 2-indents ones just make things unreadable so I don't want them to spread.

```suggestion
function usr$not(_v0) -> _result {
    switch _v0
        case false { _result := true;  leave }
        case true  { _result := false; leave }
}
```

cameel: Also, there are no semicolons in Yul. I don't think this snippet would compile.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
cameel: Similar issue with `require()` and `not()` in the section about Yul. It reads as if you're referencing some well-known functions from stdlib. But then you get to `isFinished()` and `tryWithdraw()` which clearly can't be that.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+```yul
+function usr$require(cond, msg) {
+  switch cond
+    case true  {}
+    case false { usr$myrevert(msg) }
+}
+```
+
+Empty `{}` blocks are valid Yul. The `case true` arm does nothing and falls
+through to the implicit `leave`.
```
cameel: I don't think "fall through" is the right term here. Falling through would mean that you jump to the next case. This can happen in C, but in Yul it's impossible. I'd say that we just exit the `switch` statement.

cameel: By the way, why do you think an empty block looks unusual? Note that `leave` is not a part of `switch` syntax in Yul. It exits the current function, not the `switch`. Not having it there is normal.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+function tryWithdraw(balance : uint256, amount : uint256) -> Option(uint256) {
+    match balance >= amount {
+    | False => return None;
+    | True  => return Some(balance - amount);
+    }
+}
+```
+
+`None` is `(false, <unused>)` and `Some(x)` is `(true, x)` — two stack slots.
+The function returns both:
```
cameel: I think this would be clearer without the `Option`. It seems to me that the point is to show how more complex conditions are compiled and the enum distracts from that. At first I actually missed it and thought the enum was the point.

Or, if you do want to specifically show what ADTs compile into, I'd put that in a separate example (without a `match`, which is not relevant to that).

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+smart contract bugs that Classic Solidity has no good defense against: the
+silent handling of missing or newly-added cases.
+
+The problems are structural. When a contract's logic depends on a set of
+alternatives - payment types, auction phases, token standards, vote outcomes -
+Classic Solidity provides enums and structs, but it cannot enforce that every
+function that dispatches on that type handles all cases, or that every variant
+carries the right fields and no others. Developers fill the gap with runtime
+`require` checks and a discipline of adding cases everywhere. That discipline is
+not enforced by the compiler, so it can and does fail - usually at the worst
```
cameel: Most of the text uses em dashes but this final section has ordinary ones. You should settle on one style.

BTW, the em dashes are a common telltale of LLM-produced text, which nowadays may inadvertently give people an impression of low quality. They are technically more correct, but I generally tend to avoid them.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+Yul has no algebraic types, only 256-bit words. Every sum type is therefore
+flattened onto the EVM stack as:
+
+```
+1 (tag word) + max(payload size across all constructors)
+```
+
+The tag identifies which constructor is active. The payload slots hold the
+fields of the active constructor, padded with unused words to reach the maximum
+payload size.
```
cameel, line 380: Side-question: are we planning to support packed enums eventually? Or if not, how do we see people implementing them in Core Solidity?

I'm asking, because I just ran into this in OZ repo: [`packValidationData()`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.6.1/contracts/account/utils/draft-ERC4337Utils.sol#L91-L105). Packing stuff into a single slot is very common and now that I think of it, an extra benefit of enums could be to make this pattern safer.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: [`ECDSA.RecoverError`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.6.1/contracts/utils/cryptography/ECDSA.sol#L13-L18) is an example of enum-based error handling that I found in the wild:
```solidity
    enum RecoverError {
        NoError,
        InvalidSignature,
        InvalidSignatureLength,
        InvalidSignatureS
    }
```

It's being handled with a series of `if`s in [`throwError()`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.6.1/contracts/utils/cryptography/ECDSA.sol#L273-L283), which is actually a good example of how people have to jump through hoops to associate a bit of data with an enum:

```solidity
    function _throwError(RecoverError error, bytes32 errorArg) private pure {
        if (error == RecoverError.NoError) {
            return; // no error: do nothing
        } else if (error == RecoverError.InvalidSignature) {
            revert ECDSAInvalidSignature();
        } else if (error == RecoverError.InvalidSignatureLength) {
            revert ECDSAInvalidSignatureLength(uint256(errorArg));
        } else if (error == RecoverError.InvalidSignatureS) {
            revert ECDSAInvalidSignatureS(errorArg);
        }
    }
```

In Core Solidity `errorArg` could be stored inside the enum rather than awkwardly passed in as a function argument.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+already use `require` checks and careful enum handling. What does this buy me?"
+
+Your existing practices aren't wrong, they just don't scale. Today, the
+discipline of "update every dispatch function when you add a variant" relies on
+memory, code review checklists, and audit reports. It is not enforced by the
+compiler, so it can fail. When a team member adds `ERC1155` support under
+deadline pressure and misses one function, the compiler says nothing. Pattern
+matching moves that discipline into the toolchain, where it cannot be forgotten.
+
+## The Pattern Match Compiler
```
cameel, line 244: I expected that we'll be describing work in progress, but the section gives an impression of a finished design. Is that intentional? If not, we might want to mention that this is the state as of some version X.Y.Z of solcore or something. And what might change in the future.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+**No compile-time totality guarantee.** If we add a new variant, say `ERC1155`,
+to the `PaymentType` enum, the compiler will not tell us where the dispatch
+logic needs to be updated. `processPayment` will silently do nothing for
+`ERC1155` payments. `calculateFee` will silently return 0. Every function that
+dispatches on `PaymentType` is a potential silent failure waiting to happen.
+
+**Redundant runtime validation.** Because the type system cannot express the
+constraint that, e.g., a `NATIVE` payment has no `token`, each function must
+repeat those structural checks. This is both verbose and error-prone: a
+developer might add a new function and forget to validate one of the fields.
```
cameel, line 99: Ironically, this bit seems a bit redundant :) I think this is just the consequence of invalid states being representable.

I'd mention something else: data cohesion. ADTs bundle the tag and all fields together, reducing the risk of mixing them up. This is something you can't easily guard against with `require()`.

cameel (commented):
There are things that could be improved but I found very few actual issues that need fixing. My comments below are mostly just suggestions. For the most part the article seems good enough already.

The main problem I noticed is `leave` in Yul. It seems to me that it's being treated as the equivalent of `break` in C's `switch` syntax, while it's really an equivalent of function's `return`.

As for the general structure, I actually expected an article focused on the progress with the pattern compiler but it looks like it's a more general one about pattern matching in Core Solidity. Which is fine, but if that's the intent then it could use more detail on how `match` works in the language. Currently it jumps abruptly from a high-level text about benefits of pattern matching to a low-level description of how patterns are compiled. The `match` syntax and capabilities are only explained indirectly through what is shown in the examples and it feels a bit like some gentler introduction is missing.

Overall, which suggestions to address is up to you. I pointed out what I thought as I went through it, but I am aware doing it all may be a bit too much. It's not like the article has to be perfect. We may benefit more from just publishing more frequently instead.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: Actually, maybe that's more of a use case for a packed struct. I guess in this code you don't really want the tag stored with the payload.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
cameel: I'd keep this one unresolved until the actual date is put in so that we don't forget about it.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: Why is the amount a `word` though? Wouldn't you typically use `uint` for it?

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
```diff
@@ -0,0 +1,521 @@
...
+produces:
+
+```
+Warning: Clause (True → return True) is redundant.
+  in function f
+  in match (x)
+```
+
+Unlike non-exhaustive matches, redundant-clause warnings do not prevent
+compilation: they are surfaced as warnings to assist developers during
```
cameel: "surfaced" does not sound natural in this context.

```suggestion
compilation: they are reported as warnings to assist developers during
```

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: I think it is. I personally don't find the style here particularly readable, though I actually think the `data` with its weird mixed 2-space/4-space indent is more annoying.

But any discussion about syntax/style is intentionally being postponed so that the team can focus on the design. I expect that we'll get to this at some point, because like it or not, it does affect how users perceive the language and we can't ignore it forever.

czepluch:
> As for the general structure, I actually expected an article focused on the progress with the pattern compiler but it looks like it's a more general one about pattern matching in Core Solidity.

The initial draft did both. First a high level overview of pattern matching and then the deep dive. We agreed that it was doing too much and had two different target groups in one post. Instead we decided to first do the high level one and then we can do the more technical one later. This way we have two posts to put out, and they have different target groups.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: Added the following sentences: 
```
These checks also have a runtime cost: every `require` executes on-chain and
consumes gas, whereas invariants enforced by the type system are verified at
compile time and vanish entirely from the deployed bytecode.
```

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Fixed. Replaced by category of bugs.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: > This example is fine, but I just wanted to mention that another (perhaps more idiomatic) option here would be error handling. This is the classic situation where forgetting an important case is common and can have disastrous consequences.
> 
> It would be something along these lines:
> 
> ```haskell
> data Result =
>     Ok(T)
>   | Error(uint256);
> 
> function doStuff() -> T {
>     match stuff() {
>     | Ok(result) => return result;
>     }
> }
> ```
This example will result in an error, since the match is not total over type `Result`.

rodrigogribeiro: > Though the issue I see after writing it down is that you'd be unlikely to use `match` here unless you already remember that you have to handle the error. Core Solidity has no equivalent for Rust's `#[must_use]` or C++'s `[[nodiscard]]`, does it? :)

I'm not aware of these features in Rust and C++. So, I could say that Core does not support it. :)

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: I'll rewrite all parts that use dashes into equivalent sentences.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: The post idea is just to explain the current status of the pattern matching in the language. Previously, solcore used a pattern matching compiler based on backtracking automata, which is hard to deal with totality checking. I'm not sure if is a **complete** design. My idea was to change the compiler in order to support exaustiveness checking.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: Ok, added the following: 
```There is also a
subtler issue of data cohesion. Because the `Payment` struct separates the tag
(`paymentType`) from its fields, nothing prevents a caller from pairing them
incorrectly. ADTs bundle the constructor and its payload into a single,
inseparable unit, so that kind of mismatch cannot be expressed in the type and
no `require` check can fully substitute for that guarantee.
```

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: I believe that we could make another post to show how language new constructs will enable other ways to express this kind of pattern.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Added text to explain the gas cost.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Rewritten into equivalent sentences.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: > This example will result in an error, since the match is not total over type `Result`.

Yes, that was the point. Just like your example does not cover `Cancelled` and `Ended`. I'm just questioning whether anyone would write it like his in the first place. They'd probably just call the function without doing a `match` on the result.

> I'm not aware of these features in Rust and C++.

They basically don't let you ignore the result. You have to use it, which means that if it's an enum and you don't consider some variant, the compiler has a chance to detect it.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: This would be an interesting detail to have in the article then.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: Yeah, that would be interesting to read about.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: I think this comment was meant for https://github.com/argotorg/solidity-website/pull/196#discussion_r3120410500?

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: Added text explaining the importance on changing the pattern matching compiler.

cameel:
@rodrigogribeiro Just a heads up, because I have a feeling you missed some comments from me and @czepluch :) Note that github hides some of them behind behind a box like this:
```
5 hidden conversations
Load more...
```

You did not respond to any of them while you covered pretty much all the visible ones, which makes me think you have not seen them at all.

czepluch:
Im pushing a bunch of those fixes and updating the frontmatter in about an hour so we can release tomorrow

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
czepluch: Totally agree.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: > I think this comment was meant for [#196 (comment)](https://github.com/argotorg/solidity-website/pull/196#discussion_r3120410500)?

Yes, apologies for the confusion.

rodrigogribeiro: > It would make more sense to declare `to` as `address payable` in the struct since it's an address that's meant to receive payments.

Changed field name to `to`

rodrigogribeiro: > Same for Core Solidity version actually. Does it have `address payable`?

No, we do not support record syntax yet.

> And what's `transfer()` that you use in it? A stdlib function or something user-defined? I guess there's no syntax sugar for `<address>.call()` yet?

It is a user defined function. Do you think that we need to mention it? The idea is just to show how pattern matching improves safety in the language.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: Ok, now I could say for sure that Core does not support it. :)
But, I do not see a reason to support such feature.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: I do not get your point here. The example makes sense for me.

rodrigogribeiro:
> @rodrigogribeiro Just a heads up, because I have a feeling you missed some comments from me

Sorry, @cameel... I went now for whole conversation and some of these are comments were not seen by myself. I've double checked and I believe that I covered all of your comments. The only point that I left as it is was about the example about `Option` which, at least for me, makes sense.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
cameel: Maybe you're right. My point was the `Option` is distracting because it seems like it's relevant to the example while it's not. You could just as well reduce it to something like this:

```js
function withdrawableAmount(balance : uint256, amount : uint256) -> uint256 {
    match balance >= amount {
    | False => return 0;
    | True  => return balance - amount;
    }
}
```

Though I also expected that this would somewhat reduce the generated assembly, but looking at it now, the part generated by `Option` is not that complex:

```yul
      _result_tag     := true
      _result_payload := sub(balance, amount)
```
so it does not improve as much as I thought.

cameel:
@rodrigogribeiro Have you seen https://github.com/argotorg/solidity-website/pull/196#discussion_r3120842167 and https://github.com/argotorg/solidity-website/pull/196#discussion_r3120801255?

This is my biggest problem with the sample code in the post. It seems to treat `leave` as a mandatory part of `switch` syntax for some reason. And it uses semicolons in Yul :)

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: I think this should be a consideration at some point, Not necessarily for ADTs in general but for error handling specifically. It's not like you have to have that in the language (Classic Solidity does not have an equivalent), but it's a common and useful mechanism in languages we're taking inspiration from. Both Rust and Zig have mechanisms that force you to handle errors. A replacement for exceptions that's baked into the type system instead of relying on the implicit stack unwinding. Having neither of those we're basically back to C.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
cameel: > Do you think that we need to mention it? The idea is just to show how pattern matching improves safety in the language.

Not sure. I think it may be unclear to people because they might not expect that Core Solidity does not have `call()`. But also this is not the point of the example so maybe it's still fine.

rodrigogribeiro: Ok. Since it is not the focus, I'll leave as it is now.

rodrigogribeiro:
> @rodrigogribeiro Have you seen [#196 (comment)](https://github.com/argotorg/solidity-website/pull/196#discussion_r3120842167) and [#196 (comment)](https://github.com/argotorg/solidity-website/pull/196#discussion_r3120801255)?
> 
> This is my biggest problem with the sample code in the post. It seems to treat `leave` as a mandatory part of `switch` syntax for some reason. And it uses semicolons in Yul :)

Ohhhh silly me... Fixed.

src/posts/2026-04-16-pattern-matching-in-core-solidity.md
rodrigogribeiro: Fixed.

src/posts/2026-04-27-pattern-matching-in-core-solidity.md
rodrigogribeiro: > They basically don't let you ignore the result. You have to use it, which means that if it's an enum and you don't consider some variant, the compiler has a chance to detect it.

Ok, I understood you point and I agree that this could be a must in Core Solidity. Do you mind to add this as an issue? Just to avoid forget the idea and to implement it in a not distant iteration.

cameel: We actually just had a discussion with Lexi about that on the channel and ended up with a suggestion to do just. I was going to add it tomorrow, but if you want to do it yourself, feel free, less work for me :) But I would suggest reading that discussion first and adding any relevant info from there (that's what I was going to do).

czepluch (approved):

PR #205, opened 2026-05-19
Misordered named parameters bug blog post

nikola-matic:


src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
+---
+layout: post
+published: true
+title: 'Misordered Named Parameters in require with Custom Errors Bug'
+date: '2026-05-20'
```
nikola-matic, line 5: Don't forget

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
...
+author: Solidity Team
+category: Security Alerts
+---
+
+On February 9, 2026, the Solidity team discovered a bug in the IR-based code generator that
```
nikola-matic: Do we know the actual reporter (aside from the EF bug bounty), and should we use them instead of "Solidity team"?

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
...
+causes the arguments of a custom error passed to `require` using named-parameter syntax to be
+ABI-encoded in call-site order rather than declaration order.
+
+The bug was introduced in Solidity 0.8.26, which added support for passing custom errors as
+the second argument to `require`.
+[Solidity version 0.8.36](https://github.com/ethereum/solidity/releases/tag/v0.8.36), released on May 20, 2026, provides a fix.
```
czepluch: update date when we 0.8.36

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
+---
+layout: post
+published: true
+title: 'Misordered Named Parameters in require with Custom Errors Bug'
```
czepluch, line 4:
```suggestion
title: 'Misordered Named Parameters in `require` with Custom Errors Bug'
```

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
czepluch: Seems like a fairly long time to fix. Maybe worth mentioning in the next paragraph that due to the "very low" severity, it was not prioritized until now.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
czepluch:
```suggestion
[Solidity 0.8.36](https://github.com/ethereum/solidity/releases/tag/v0.8.36), released on May 20, 2026, provides a fix.
```
In the same paragraph, we say "Solidity 0.8.26", so let's be consistent. Past posts also omit "version"

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
...
+[Solidity version 0.8.36](https://github.com/ethereum/solidity/releases/tag/v0.8.36), released on May 20, 2026, provides a fix.
+
+We assigned the bug a severity of "very low".
+The affected code is the construction of revert data for a transaction that is already going
+to revert, so contract state is not modified by it.
+Triggering the bug additionally requires a fairly specific pattern - a custom error passed to
+`require` and instantiated with named arguments in a non-declaration order - and only affects
+the IR-based pipeline.
+
+## Which Contracts are Affected?
```
czepluch:
```suggestion
## Which Contracts Are Affected?
```
previous posts use a header style with all words capitalized.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
...
+reordering them to match the parameter order of the error. With the snippet above, the
+contract reverts with a payload whose first word holds the value of `b` (`7`) and whose
+second word holds the value of `a` (`2`), so off-chain decoders read `a = 7, b = 2` instead
+of the intended `a = 2, b = 7`.
+
+Standalone `revert ErrorName({...})` statements go through a different code path that
+reorders arguments before encoding, and so are not affected. The bug is specific to the
+`require`-with-custom-error path introduced in 0.8.26.
+
+### Misalignment with Mixed Stack Widths
```
czepluch:
```suggestion
### Misalignment With Mixed Stack Widths
```
previous posts use a header style with all words capitalized.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
...
+3. The named arguments are written in an order that differs from the order declared by the
+   error definition. Positional arguments, or named arguments written in declaration order,
+   produce the correct encoding.
+
+Other language constructs that accept named arguments are **not** affected, including:
+
+- `revert ErrorName({...})` statements,
+- event emissions (`emit E({...})`),
+- internal, external, library, and bound function calls,
+- struct constructor invocations.
```
czepluch:
```suggestion
- `revert ErrorName({...})` statements
- event emissions (`emit E({...})`)
- internal, external, library, and bound function calls
- struct constructor invocations
```
trailing commas on a bullet list reads weird, let's just remove trailing punctuation

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
...
+
```
czepluch, line 54: Add something that explicitly states a workaround until upgrading:
> Until upgrading, the bug can be avoided by writing the named arguments in declaration order, or by using positional arguments.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,126 @@
...
+```solidity
+require(false, NamedArgsError({b: 7, a: 2}));
+```
+
+The compiler binds named arguments to parameters *by name*, so any permutation that mentions
+every parameter exactly once is accepted by the type checker. The type system does not
+require the call-site order to match the declaration. This means the bug can be triggered by
+any reordering - including reorderings of two arguments of the same type, where no positional
+type mismatch could ever surface - which was the most surprising aspect of the bug.
```
czepluch: It's a bit hard to read this sentence, since there's a comma inside the parenthetical dashes. Consider restructuring.

czepluch (changes requested):
looks good. mostly cosmetic nits from my side.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
nikola-matic: This will just render with backticks instead of a "code sections", so can't really apply this one :)

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+---
+
+On February 9, 2026, a bug in the IR-based code generator was reported through the Ethereum
+Foundation bug bounty program. The bug causes the arguments of a custom error passed to
+`require` using named-parameter syntax to be ABI-encoded in call-site order rather than
+declaration order.
+
+The bug was introduced in Solidity 0.8.26, which added support for passing custom errors as
+the second argument to `require`.
+[Solidity 0.8.37](https://github.com/ethereum/solidity/releases/tag/v0.8.37), released on May 20, 2026, provides a fix.
```
cameel: This date needs to be updated too, but I'd actually just omit it. Specifying version is unambiguous enough.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+Due to the very low severity of the bug, described below, the fix was not prioritized for an
+expedited release.
+
+We assigned the bug a severity of "very low".
+The affected code is the construction of revert data for a transaction that is already going
+to revert, so contract state is not modified by it.
```
cameel:
```suggestion
to revert, so it has no impact on contract's state.
```

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+- event emissions (`emit E({...})`)
+- internal, external, library, and bound function calls
+- struct constructor invocations
+
+Until upgrading to a fixed version, the bug can be avoided by writing the named arguments in
+declaration order, or by using positional arguments.
+
+## Technical Details
+
+A custom error declaration fixes the order of its parameters, and that order determines its
```
cameel:
```suggestion
A custom error declaration specifies the order of its parameters, and that order determines its
```

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+ABI signature and the layout of its encoded payload:
+
+```solidity
+error NamedArgsError(uint256 a, uint256 b);
+```
+
+Solidity also allows the caller to pass arguments by name, optionally in a different order
+from the declaration:
```
cameel: "also" here sounds weird, as if you just showed the version with positional arguments, but you did not.

```suggestion
Solidity allows the caller to pass arguments by name, optionally in a different order
from the declaration:
```

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+
+```solidity
+require(false, NamedArgsError({b: 7, a: 2}));
+```
+
+The compiler binds named arguments to parameters *by name*, so any permutation that mentions
+every parameter exactly once is accepted by the type checker. The type system does not
+require the call-site order to match the declaration. As a result, the bug can be triggered
+by any reordering. Most surprisingly, this includes reorderings of two arguments of the same
+type, where no positional type mismatch could ever surface.
```
cameel: Why is that surprising? To me it's actually the more obvious case. Lack of type mismatch means that the compiler or decoder can't detect it so no compilation or runtime error can save you. It's the fact that the compiler won't fail trying to encode mismatched types that was more surprising to me.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+
+In the IR-based code generator, the lowering of `require` with a custom error passes the
+arguments to the encoding helper in the order they appear at the call site, instead of first
+reordering them to match the parameter order of the error. With the snippet above, the
+contract reverts with a payload whose first word holds the value of `b` (`7`) and whose
+second word holds the value of `a` (`2`), so off-chain decoders read `a = 7, b = 2` instead
+of the intended `a = 2, b = 7`.
```
cameel: This will be fixed at the time the post go out so it should be written in past tense.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+
+Standalone `revert ErrorName({...})` statements go through a different code path that
+reorders arguments before encoding, and so are not affected. The bug is specific to the
+`require`-with-custom-error path introduced in 0.8.26.
+
+### Misalignment With Mixed Stack Widths
+
+The consequences extend beyond a swap of word-sized values when the arguments occupy
+different numbers of stack slots in the Yul IR. References to `string`, `bytes`, and arrays
+in `calldata` are represented by *two* stack slots (offset and length), whereas value types
```
cameel: - This is redundant. `string` and `byte` are arrays. It would be more relevant to mention slices instead.
- This is not Yul-specific. The bug itself was, but the the stack layout is the same between codegens.
- Some value types also use two stack slots (external function pointers).

```suggestion
different numbers of stack slots. References to arrays and array slices
in `calldata` are represented by *two* stack slots (offset and length), whereas most value types
```

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+and references in other data locations use a single slot. When a `calldata` reference is
+reordered with respect to a value-type argument, the slots passed to the encoding helper are
+permuted at the level of *stack slots*, not at the level of named parameters, and the helper
+interprets its inputs against parameter positions of a different width than intended.
```
cameel: They are not permuted as individual stack slots. They're still grouped. Just in the wrong way.

Also, "parameter positions of a different width" sounds weird to me. I know what you mean, but only because I already know how the bug works. Generally, referring to it as "width" or "stack width" sounds wrong.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+```
+
+The encoding helper for `StringAndUint` expects to first receive the two slots that describe
+`a` (the calldata string's offset and length), followed by the single slot that holds `b`.
+With the bug it receives `b`'s single slot first, followed by `a`'s two slots, and then
+encodes them assuming the first two slots belong to the dynamic-type parameter. The length
+field is therefore read from a value-type slot, and the string data is copied from an
+arbitrary calldata offset.
```
cameel:
```suggestion
field is therefore read from a value-type slot, and the string data is copied from a
random calldata offset.
```

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+
+In the most severe configurations the encoding aborts with an EVM-level error, which still
+causes the transaction to revert - but with no decodable error data - rather than producing
+the intended custom-error revert. In less severe configurations the payload is well-formed
```
cameel: I'd say that the latter is worse, because it's sometimes possible for the receiver to successfully decode it.

Actually, the current text makes it sound like the data can never be decoded. Often it actually can. It's just numbers. A small integer can easily pass for a valid offset.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+reordering them to match the parameter order of the error. With the snippet above, the
+contract reverts with a payload whose first word holds the value of `b` (`7`) and whose
+second word holds the value of `a` (`2`), so off-chain decoders read `a = 7, b = 2` instead
+of the intended `a = 2, b = 7`.
+
+Standalone `revert ErrorName({...})` statements go through a different code path that
+reorders arguments before encoding, and so are not affected. The bug is specific to the
+`require`-with-custom-error path introduced in 0.8.26.
+
+### Misalignment With Mixed Stack Widths
```
cameel: You mention two cases in the post:
- swapping parameters of the same type
- swapping parameters of different types represented using different numbers of stack slots

The third possibility is missing: different types with same numbers of stack slots. It's also not considered in the impact section.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,132 @@
...
+calls. Its primary consequence is misleading off-chain consumers - block explorers, indexers,
+error-decoding libraries, and test frameworks - that decode the payload as if the values had
+been placed correctly. In the stack-misalignment cases the failure is more visible: an
+encoding-level revert can replace the intended custom-error revert, so off-chain code that
+distinguishes specific custom errors from generic reverts may take a different branch than
+expected.
+
+Named-argument syntax is uncommon in `require` calls in practice, and the bug remained
+undetected for almost two years. The narrow trigger conditions and the read-only nature of the
+affected code path together place this bug in the "very low" severity tier.
```
cameel: - It can have on-chain effects - reverts can be intercepted with try/catch or by using a bare call. It's just that this is not (and should not be) commonly done.
- Do we want to also mention the ERC that made us reconsider the severity of this?

cameel (commented):
Just wording and minor inaccuracies. The overall structure is fine.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
cameel: It was people from Spearbit. Hari will know.

Or you can ping EF - we still did not notify them that we changed the severity of this after our initial response and did not send impact analysis.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,167 @@
...
+We assigned the bug a severity of "very low".
+The affected code is the construction of revert data for a transaction that is already going
+to revert, so it has no impact on contract's state.
+Triggering the bug additionally requires a fairly specific pattern - a custom error passed to
+`require` and instantiated with named arguments in a non-declaration order - and only affects
+the IR-based pipeline.
+
+## Which Contracts Are Affected?
+
+A contract is affected when **all** of the following conditions hold:
```
clonker, line 30: string/bytes literals are inserted directly at declaration order, so these are unaffected as far as i can tell - then you'd need at least two non-string/byte lit named args

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,167 @@
...
+    function f(string calldata s) external pure {
+        require(false, StringAndUint({b: 42, a: s}));
+    }
+}
+```
+
+The encoding helper for `StringAndUint` expects to first receive the two slots that describe
+`a` (the calldata string's offset and length), followed by the single slot that holds `b`.
+With the bug it receives `b`'s single slot first, followed by `a`'s two slots, and then
+encodes them assuming the first two slots belong to the dynamic-type parameter. The length
```
clonker: the params are `(a_offset, a_length, b)` and the call site yields a permutation `(42, s_offset, s_length)`, so the offset becomes `42`, the length becomes the offset, and the value of `b` is the length, isn't it? So it would be that the offset field is read from a value type slot, not the length.

clonker (commented):
Some small comments. Should we mention that the `revert` always happens? (it does, doesn't it? just with weird data).

cameel:
> Should we mention that the `revert` always happens? (it does, doesn't it? just with weird data).

It does say that already, in a way:
>> The affected code is the construction of revert data for a transaction that is already going to revert, so it has no impact on contract's state.

>> Because the bug affects only the construction of revert data for transactions that are already going to revert, it does not modify storage, return values, or external calls in the reverting transaction itself.

I wanted to suggest that too at first but in the end decided that the above are already enough. But if you also thought that, maybe there's something to it and we should make that more explicit.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
cameel:
```suggestion
date: '2026-09-10'
```

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,167 @@
...
+instead of the intended `a = 2, b = 7`.
+
+Standalone `revert ErrorName({...})` statements go through a different code path that
+reorders arguments before encoding, and so were not affected. The bug was specific to the
+`require`-with-custom-error path introduced in 0.8.26.
+
+Reordering parameters of *different* types that occupy the same number of stack slots
+(say, a `uint256` and an `address`) also produces no misalignment: each value simply lands
+in the other parameter's position and is encoded as if it had the other type. The payload
+stays structurally valid, but decoders reinterpret the values under the wrong types.
```
cameel: This should be in the past tense too.

```suggestion
Reordering parameters of *different* types that occupy the same number of stack slots
(say, a `uint256` and an `address`) also produced no misalignment: each value simply landed
in the other parameter's position and was encoded as if it had the other type. The payload
stayed structurally valid, but decoders reinterpreted the values as the wrong types.
```

And the section below as well.

src/posts/2026-05-20-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,167 @@
...
+words are just numbers, and a small integer can easily pass for a valid offset. This case is
+arguably worse - a consumer that matches on the error selector and decodes the fields
+observes plausible-looking values that have no relation to the source-level arguments.
+
+## Impact
+
+Because the bug affects only the construction of revert data for transactions that are
+already going to revert, it does not modify storage, return values, or external calls in the
+reverting transaction itself. Revert data can, however, be observed on-chain: a caller may
+intercept it with `try`/`catch` or a bare `call` and branch on its contents, although doing
```
cameel: I said "bare call" but that's only our internal terminology. Not sure why, but it's never called that in the official docs. So I'd refer to it like this instead:

```suggestion
intercept it with `try`/`catch` or as a result of a low-level `CALL` and branch on its contents, although doing
```

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
cameel: Why would they be unaffected? They're still a part of the encoding.

I wanted to point at `semanticTests/errors/require_error_named_parameters.sol` to prove that, but looks like the fix PR adds new tests after the fix so its effects are not visible in the commit.

@nikola-matic could you separate them?

clonker: I will have to look at that again, it's been a month :)

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+bug causes the arguments of a custom error passed to `require` using named-parameter syntax
+to be ABI-encoded in call-site order rather than declaration order.
+
+The bug was introduced in Solidity 0.8.26, which added support for passing custom errors as
+the second argument to `require`.
+[Solidity 0.8.37](https://github.com/ethereum/solidity/releases/tag/v0.8.37) provides a fix.
+Due to the very low severity of the bug, described below, the fix was not prioritized for an
+expedited release.
+
+We assigned the bug a severity of "very low".
```
clonker: I don't love the repetition of "very low" here but it's also fine to keep it.


```suggestion
We assigned the bug a severity of "very low".
Therefore, the fix was not prioritized for an expedited release.
```

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+diligence for the fix, all of them were checked and covered with tests:
+
+- `revert ErrorName({...})` statements
+- event emissions (`emit E({...})`)
+- internal, external, library, bound, and virtual function calls, including calls via function pointers
+- getters
+- struct constructor invocations
+- inherited constructors
+- modifier invocations
+- type conversions
```
clonker: do type conversions actually take named arguments? In case they don't, you could rephrase it as


```suggestion
Other language constructs that take similar code paths are **not** affected. As part of the due
diligence for the fix, all of them were checked and covered with tests:

- `revert ErrorName({...})` statements
- event emissions (`emit E({...})`)
- internal, external, library, bound, and virtual function calls, including calls via function pointers
- getters
- struct constructor invocations
- inherited constructors
- modifier invocations
- type conversions
```

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+
+Until upgrading to a fixed version, the bug can be avoided by writing the named arguments in
+declaration order, or by using positional arguments.
```
clonker: You could put that and a note to upgrade `solc` into a `## Recommended Actions` section (which we usually have for such blogposts, I think)

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+
+Depending on the actual values, the encoding could abort with an EVM-level error, which still
+caused the transaction to revert - but with no decodable error data - rather than producing
+the intended custom-error revert. Alternatively, the contract reverted with a payload that
+did not match the error's ABI signature but could still decode successfully: the misplaced
+words are just numbers, and a small integer can easily pass for a valid offset. This case is
+arguably worse - a consumer that matches on the error selector and decodes the fields
+observes plausible-looking values that have no relation to the source-level arguments.
+
+## Impact
```
clonker:
```suggestion
## Severity Assessment
```

if you'd like to be consistent with other bug blog posts

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
clonker: They're filtered out from the arguments list is why:

https://github.com/argotorg/solidity/blob/72df0e5652e78e42747a186e5098f40b6847ea0c/libsolidity/codegen/ir/IRGeneratorForStatements.cpp#L1173

string and bytes literals have size on stack 0

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+
+A contract is affected when **all** of the following conditions hold:
+
+1. The contract is compiled with the IR-based pipeline (`--via-ir` on the command line or
+   `viaIR: true` in Standard JSON). The evmasm pipeline is unaffected.
+2. The contract uses `require(condition, ErrorName({...}))` where `ErrorName` is a user-defined
+   error.
+3. The named arguments are written in an order that differs from the order declared by the
+   error definition. Positional arguments, or named arguments written in declaration order,
+   produce the correct encoding.
```
clonker:
```suggestion
3. At least two arguments that are not string or bytes literals appear in an order differing from the declaration.
  Positional arguments, or named arguments written in declaration order, produce the correct encoding.
```

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+reverting transaction itself. Revert data can, however, be observed on-chain: a caller may
+intercept it with `try`/`catch` or as a result of a low-level `CALL` and branch on its contents, although doing
+so is uncommon. The primary consequence is misleading consumers of revert data - block
+explorers, indexers, error-decoding libraries, and test frameworks - that decode the payload
+as if the values had been placed correctly, observing swapped or type-confused values. In
+the stack-misalignment cases the failure is more visible: an encoding-level revert can
+replace the intended custom-error revert, so code that distinguishes specific custom errors
+from generic reverts may take a different branch than expected.
+
+Revert data can also be load-bearing:
```
clonker: nice claudism lol

clonker (commented):
i think the trigger condition should be addressed (the comment thread btwn kamil and me)

beyond that the post diverges a bit from what we usually have as structure in bugfix posts. but maybe that's alright, i have no strong opinion. i think it would benefit from a `Affected versions` sub section, though.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
nikola-matic: Yeah, reads more nicely. I think merging the two sentences would be even better, i.e.

We assigned the bug a severity of "very low", therefore, the fix was not prioiritized for an expedited release.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
nikola-matic: @clonker's correct on this one, zero stack params are encoded properly both before and after the fix.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
nikola-matic: Ha, you're technically correct I think - I write/edit all markdown files in Clion and it keeps underlining every potentially 'compound' statement as wrong and suggests adding a dash in between.

nikola-matic: Unless you're referring to load-bearing, in which case you're definitely correct :D

clonker: yep, that was the load-bearing hint :D

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
nikola-matic: They don't - I believe that Kamil requested to double check all of these just in case, which is why I put it there. I would have preferred to have this kept strictly to named parameters, but considering that I also applied the `arguments()` -> `sortedArguments()` change to other places, it does make sense to use your suggestion.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
nikola-matic: >   Positional arguments, or named arguments written in declaration order, produce the correct encoding.

Doesn't really fit given the context of the section (Which contracts are affected?).

clonker (approved):
lgtm!

czepluch (approved):

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
cameel: Oh, I only skimmed this list and missed the fact that it also includes things that do not support named args. I'd drop those. [What I suggested in the PR](https://github.com/argotorg/solidity/pull/16453#discussion_r2942406841), was to make sure that the fact that they do not accept named args has test coverage, but I don't think they're especially relevant here.

I mean, it's fine to keep them with @clonker's adjustment, but the main point of this bit to me seems to be just to reassure the user that they can safely use other constructs that support named args. TBH I don't think we even need to list them. We could just say we checked them all.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
cameel: This was marked as resolved, but looks like you simply added `Recommended Actions` with a note to upgrade, but did not move this bit as @clonker requested.

BTW we only started including such a section recently, and I have mixed feelings about it. In a case like this where we can actually recommend a workaround, it may be useful. But in cases where we can only say obvious things (upgrade to the next version) it's just pointless padding. Maybe it would make more sense to have a section specifically about possible workarounds (when they are possible).

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
cameel: Ah, right. I was confused because I still default to thinking about it as a reordering of ABI-encoded representations of parameters rather than a swap of items on the stack before encoding.

Actually, I think it's to some extent a problem with the post itself though. It does not make that distinction clear. I had to get back to this bug several times now and I always start off thinking about it this way until I realize that it's the wrong way to look at it. I suspect that this will be the default mental model for users too when they read about misordering of error parameters.

The technical details should really explain that the parameters go through different layers (invocation in the source, internal representation on the stack, ABI-encoded representation in returndata) and that the problem is on the second rather than the third one.

Currently there are a few bit in the post that actually reinforce the wrong notion. For example, the post starts off with this sentence, which technically can be interpreted either way, but really sounds more like a reordering of the encodings:

> causes the arguments of a custom error passed to `require` using named-parameter syntax to be ABI-encoded in call-site order rather than declaration order

Another sentence in the technical details says this:

> The payload stayed structurally valid, but decoders reinterpreted the values as the wrong types.

which sounds very much as if the decoder was e.g. taking an `uint` encoded as `uint` and wrongly assuming it needs to be decoded as if it was `bytes`. Instead it's really the encoder doing the reinterpreting here. It treats the `uint` as a memory offset, grabbing some random data and producing an actual `bytes` value from it that the decoder simply reads.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
nikola-matic: I'd actually prefer to remove the list.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+title: 'Misordered Named Parameters in require with Custom Errors Bug'
+date: '2026-09-10'
+author: Solidity Team
+category: Security Alerts
+---
+
+On February 4, 2026, a bug in the IR-based code generator was reported by Carl from
+[Spearbit](https://spearbit.com/) through the Ethereum Foundation bug bounty program. The
+bug causes the arguments of a custom error passed to `require` using named-parameter syntax
+to be ABI-encoded in call-site order rather than declaration order.
```
cameel:
```suggestion
bug caused the arguments of a custom error passed to `require` using named-parameter syntax
to be passed to the ABI encoder in call-site order rather than declaration order.
```

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Which Contracts Are Affected?
+
+A contract is affected when **all** of the following conditions hold:
+
+1. The contract is compiled with the IR-based pipeline (`--via-ir` on the command line or
+   `viaIR: true` in Standard JSON). The evmasm pipeline is unaffected.
+2. The contract uses `require(condition, ErrorName({...}))` where `ErrorName` is a user-defined
+   error.
+3. At least two arguments that are not string or bytes literals appear in an order differing from the declaration.
```
cameel: For clarity I'd make the types a separate condition. We should also explain later in the post why string literals were not affected. Currently the post fails to do that.

Also, we don't really have distinct `string` and `bytes` literals. We instead have various kinds of literals that are convertible to either of those (ordinary strings, unicode strings, hex strings). I'd refer to them collectively as string literals.

```suggestion
3. At least two arguments appear in an order differing from the declaration.
4. At least one of the reordered arguments is not a string literal.
```

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+nothing can catch the mistake, since there is no type mismatch for the compiler or a decoder
+to detect. Perhaps more surprisingly, the compiler would also silently encode reorderings of
+arguments of *different* types, without any error during compilation.
+
+In the IR-based code generator, the lowering of `require` with a custom error passed the
+arguments to the encoding helper in the order they appeared at the call site, instead of
+first reordering them to match the parameter order of the error. With the snippet above, the
+contract reverted with a payload whose first word held the value of `b` (`7`) and whose
+second word held the value of `a` (`2`), so off-chain decoders would read `a = 7, b = 2`
+instead of the intended `a = 2, b = 7`.
```
cameel: I think that we should insert an extra paragraph here that gives the necessary background on how the data flows from the call to the encoding:
1. The user instantiates a error specifying the arguments in arbitrary order.
2. The compiler emits code that puts them on the stack in the declaration order.
3. The ABI encoder walks the declaration. For each declared parameter it takes the corresponding argument from the top of the stack and appends an encoded representation of the parameter to the end of returndata.

And give more detail on the stack representation:
- Most types simply require one slot, which may hold the value (for value types) or a reference to the value in memory, storage, etc. (reference types)
- Some value types (external function pointers), require more than one slot to hold the value
- Some reference types (dynamic calldata arrays), require an extra slot to hold the length
- Some literals (string literals) are not passed on the stack because the value is a part of the type (effectively, it gets hard-coded in the helper).

And finally, that arguments on the stack are simply an amorphous sequence of slots. Values are untyped and nothing enforces boundaries between them.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+
+Standalone `revert ErrorName({...})` statements go through a different code path that
+reorders arguments before encoding, and so were not affected. The bug was specific to the
+`require`-with-custom-error path introduced in 0.8.26.
+
+Reordering parameters of *different* types that occupy the same number of stack slots
+(say, a `uint256` and an `address`) also produced no misalignment: each value simply landed
+in the other parameter's position and was encoded as if it had the other type. The payload
+stayed structurally valid, but decoders reinterpreted the values as the wrong types.
```
cameel: To clarify that the reinterpretation happens during encoding, not decoding:

```suggestion
in the other parameter's position and was misinterpreted by the encoder as if it had the other type.
The resulting encoding stayed structurally valid and matched the error definition, but decoding it would not reproduce the values originally passed into the call.
```

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Misalignment With Multi-Slot Arguments
+
+The consequences extended beyond a swap of word-sized values when the arguments occupied
+different numbers of stack slots. References to arrays and array slices
+in `calldata` are represented by *two* stack slots (offset and length), whereas most value types
+and references in other data locations use a single slot. When a `calldata` reference was
```
cameel: This makes it sound as if the single-slot case can only swap value types and it's only the multi-slot one that mangles references.

Misalignment does affect only multi-slot types, but misinterpretation of offsets can affect both. You can have e.g. a memory array vs an integer. Or memory array vs storage array. Or one struct type vs another one with different fields. I think these should be described as two different problems, each with its own subsection instead of both being lumped together as misalignment.

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+encoded them assuming the first two slots belonged to the dynamic-type parameter. The offset
+field was therefore read from a value-type slot (`42`), the length field from what was actually
+the string's calldata offset, and `b` ended up holding the string's length. The string data was
+thus copied from an unrelated calldata offset, with a length that had nothing to do with the
+string.
+
+Depending on the actual values, the encoding could abort with an EVM-level error, which still
+caused the transaction to revert - but with no decodable error data - rather than producing
+the intended custom-error revert. Alternatively, the contract reverted with a payload that
+did not match the error's ABI signature but could still decode successfully: the misplaced
```
cameel: This is wrong. The payload did always match the signature. The values would be different, but if the signature said it's `(uint, string)`, you would still get an encoding of `(uint, string)`, not `(string, uint)` or something else. Values going into it would be different, but the "shape" would not change.

cameel: Actually, it's wrong in the bug list entry in https://github.com/argotorg/solidity/pull/16453 as well. I did not catch it there, but it says things like:

> producing a payload of an entirely different shape than the error's ABI signature

src/posts/2026-09-10-misordered-named-parameters-in-require-with-custom-errors-bug.md
```diff
@@ -0,0 +1,169 @@
...
+the stack-misalignment cases the failure is more visible: an encoding-level revert can
+replace the intended custom-error revert, so code that distinguishes specific custom errors
+from generic reverts may take a different branch than expected.
+
+For exmaple, [ERC-3668 (CCIP Read)](https://eips.ethereum.org/EIPS/eip-3668) relies on clients decoding
+the `OffchainLookup` custom error to drive off-chain data retrieval. While ERC-3668 includes
+mechanisms that protect against forged revert errors, they do not help when an honest
+contract emits mis-encoded revert data due to a compiler bug. Since `OffchainLookup` mixes
+dynamic and value types, a misordering in an affected `require` would most likely produce a
+garbled payload that breaks the lookup flow, though not necessarily in an obviously detectable
```
cameel: Same here. I think that there's actually no case where the data could not be decoded. It will not decode back to the original values, but if encoding does not revert, I don't think it can produce something with that's invalid due to not having enough items, having wrong lengths, leaving dirty padding, etc.

cameel (commented):
Here are more concrete suggestions on how to address https://github.com/argotorg/solidity-website/pull/205#discussion_r3971420844.

Reading the post again, I realized that I missed some things because somehow I held that incorrect assumption that the bug must be affecting the structure of the encoded data.

This even crept into the bug list entry, which we'll need to correct too.

PR #208, opened 2026-07-09
Add inheritance order reversal on storage end warning bug blogpost

clonker:


cameel (approved):
I'm assuming this is identical to what I just reviewed in the security repo so approving with the same rationale :)

PR #209, opened 2026-07-09
add intersection recursive cycles blogpost

clonker:


cameel (approved):
I skimmed through it and I'd have some remarks for the sections that were added since my last review (especially with the affected versions, we may have gone a bit overboard by including absolutely everything from my analysis rather than just the important bits :)), but overall there's nothing broken and the content is still solid so I think we can merge it already and improve later. I'll probably submit a follow-up PR with some improvements, but at least we can consider the release done :)

src/posts/2026-07-09-unsound-spill-in-mutual-recursion-bug.md
```diff
@@ -0,0 +1,231 @@
+---
+layout: post
+published: true
+title: "Unsound Spill In Mutual Recursion Bug"
+date: "2026-07-08"
```
cameel:
```suggestion
date: "2026-07-09"
```

nikola-matic:
```suggestion
date: "2026-07-09"
```

PR #213, opened 2026-09-09
Spill slot collision blogpost

czepluch:


src/styles/MDStyles.tsx
cameel, line 1: This should go into a separate PR and be merged already.

public/videos/2026/spill-slot-collision-explainer.mp4
cameel, line 1: I was afraid that the video would be huge, but this actually seems acceptable for a one-off.

Still, would probably be nicer to have the sources committed. How bulky are they?

clonker: Not super bulky. It's one python script. I can PR it next week.

src/styles/MDStyles.tsx
czepluch: good point. separate PR created in #215

public/videos/2026/spill-slot-collision-explainer.mp4
czepluch: i thought 6mb is okay to add directly to the repo. Had it been above 10mb I would have hosted it somewhere.

cameel: Yeah, perfectly fine. I'm just thinking about how it will scale, because the video is nice and I can see us wanting to do more of them :)

clonker (approved):

PR #216, opened 2026-09-10
Add 0.8.37 release announcement

czepluch:
Adds the release blog post for Solidity 0.8.37 and updates the homepage blurb to point to it.

### DIFF
--- a/src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
+++ b/src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
@@ -1,0 +1,144 @@
+---
+title: "Solidity 0.8.37 Release Announcement"
+date: "2026-09-10"
+author: Solidity Team
+category: Releases
+---
+
+We have released [Solidity v0.8.37](https://github.com/argotorg/solidity/releases/tag/v0.8.37).
+
+This is primarily a bugfix release.
+It contains two important bugfixes, both rated low/medium severity, as well as fixes for several further bugs of lower severity, three of which are described below.
+On the experimental side, the SSA CFG code generator gets a new stack shuffler that replaces the previous greedy algorithm with a planning one.
+The release also adds support for the `SLOTNUM` opcode of the Amsterdam EVM version, removes two experimental features and adds deprecation warnings for several features planned for removal.
+
+## Important Bugfixes
+
+### Memory Byte Array Element Delete Clears 32 Bytes
+
+When `delete` was applied to an element of a `bytes` array in memory, the evmasm pipeline wrote 32 zero bytes starting at that element instead of clearing only the element's byte.
+The extra zero bytes overwrote up to 31 bytes following the element.
+Those bytes may belong to later elements of the same array or, when the element sits near the end of the array, to whatever is allocated after it.
+The same applies to `string` values converted to `bytes`.
+The IR pipeline is not affected, and the analogous assignment `b[i] = 0` clears exactly one byte on both pipelines.
+All versions up to and including 0.8.36 are affected.
+
+The bug was reported by [shaheenfazim](https://github.com/shaheenfazim) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Memory Byte Array Element Delete Clears Whole Word Bug](https://blog.soliditylang.org/2026/09/10/memory-byte-array-element-delete-clears-whole-word-bug/)
+
+### Spill Slot Collision Across Mutual Recursion
+
+The stack-to-memory mover of the IR pipeline works around the EVM's stack depth limit by relocating some local variables to fixed memory offsets, called spill slots.
+When the call graph contained mutual recursion, too few slots were reserved, so variables of two different functions could be assigned the same slot even though both could be live at the same time.
+One variable was then silently overwritten by the other.
+The bug is closely related to, but distinct from, the [Unsound Spill In Mutual Recursion Bug](https://blog.soliditylang.org/2026/07/09/unsound-spill-in-mutual-recursion-bug/) fixed in 0.8.36.
+That fix does not cover this case.
+Versions 0.7.2 through 0.8.36 are affected when compiling with `--via-ir`.
+
+The bug was reported by [Ng Sze Hon](https://github.com/gzeoneth) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Spill Slot Collision Across Mutual Recursion Bug](https://blog.soliditylang.org/2026/09/10/spill-slot-collision-across-mutual-recursion-bug/)
+
+## Other Bugfixes
+
+### Misordered Named Parameters in require with Custom Errors
+
+In the IR pipeline, the arguments of a custom error constructed with named arguments and passed to `require`, as in `require(condition, MyError({b: 2, a: 1}))`, were ABI-encoded in call-site order rather than declaration order.
+Plain `revert MyError({...})` statements are not affected.
+Only the revert data of a transaction that is already reverting is affected, so the bug has no impact on contract state.
+
+The bug was reported by Carl from [Spearbit](https://spearbit.com/) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Misordered Named Parameters in require with Custom Errors Bug](https://blog.soliditylang.org/2026/09/10/misordered-named-parameters-in-require-with-custom-errors-bug/)
+
+### Uninitialized Function Pointers in Packed Storage
+
+In the evmasm pipeline, reading an uninitialized internal function pointer from a packed storage slot yielded the wrong value when a variable stored after it in the same slot held a non-zero value.
+As a consequence, an equality comparison between such a pointer and a default-initialized local function pointer could be false, even though neither had ever been assigned.
+
+### Constants Read in Both Checked and Unchecked Contexts
+
+In the IR pipeline, the code computing the value of a constant was generated once per contract, with either checked or unchecked arithmetic depending on which use site was generated first, and every other use site reused that code.
+For a constant whose initializer overflows its type, a read in a checked context could therefore miss its required overflow panic, or a read in an `unchecked` context could revert with a spurious one.
+
+## Notable Features and Changes
+
+### SSA CFG Planning Stack Shuffler (Experimental)
+
+The [experimental SSA CFG code generator](https://soliditylang.org/blog/2026/04/29/solidity-0.8.35-release-announcement/#experimental-ssa-cfg-code-generator) precomputes a stack layout for every point in the program.
+Its stack shuffler emits the `SWAP`, `DUP`, `POP` and `PUSH` instructions that get the stack from one layout to the next.
+Until now the shuffler worked greedily, applying one local repair step at a time until the target layout was reached.
+0.8.37 replaces it with a planning shuffler.
+The new shuffler first computes an explicit plan that maps the items on the current stack to their positions in the target layout, and then realizes that plan bottom-up.
+If it gets stuck because an item is out of reach for `SWAP` or `DUP`, it drops the items above the blocking one and starts over from a new plan.
+The more structured approach is easier to maintain, and initial benchmarks show it to be computationally more efficient than the greedy one while producing bytecode of roughly the same size.
+
+The SSA CFG code generator remains experimental and is enabled with `--experimental --via-ssa-cfg`.
+
+### Support for the SLOTNUM Opcode
+
+The Amsterdam EVM version, supported since 0.8.36, introduces the `SLOTNUM` opcode ([EIP-7843](https://eips.ethereum.org/EIPS/eip-7843)), which returns the beacon chain slot number of the current block.
+0.8.37 exposes it in Solidity as `block.slotnum` of type `uint64`, and in Yul and inline assembly as the builtin `slotnum()`.
+The SMTChecker supports the new member as well.
+`block.slotnum` and `slotnum()` are available only when compiling with `--evm-version amsterdam` or a later EVM version.
+From Amsterdam on, `slotnum` is also a reserved identifier in Yul.
+
+### Removals and Deprecations
+
+Two experimental features that were restricted to the [experimental mode](https://docs.soliditylang.org/en/v0.8.37/using-the-compiler.html#experimental-mode) in 0.8.35 are now removed entirely: the Language Server Protocol mode (`--lsp`) and the Generic Solidity prototype (`pragma experimental solidity`).
+
+Continuing the [deprecation work started in 0.8.31](https://soliditylang.org/blog/2025/12/03/solidity-0.8.31-release-announcement/#feature-deprecations), this release adds deprecation warnings for the EVM versions `constantinople`, `petersburg`, `istanbul` and `berlin`, and for the SMTChecker's BMC engine.
+
+## Full Changelog
+
+### Important Bugfixes
+
+* Code Generator: Fix `delete` applied to an element of a `bytes` array in memory zeroing the whole 32-byte word starting at the element's location instead of only the element's byte.
+* Yul Optimizer: Fix too few memory slots being reserved when moving local variables to memory to work around stack-too-deep errors in code containing mutually recursive functions. Variables of two functions that can be active at the same time could be assigned the same slot, silently overwriting one of them while it was still in use.
+
+### Language Features
+
+* Custom Storage Layout: Allow signed positive expressions.
+* EVM: Support `block.slotnum` to access the beacon chain slot number of the current block, available since the Amsterdam EVM version ([EIP-7843](https://eips.ethereum.org/EIPS/eip-7843)).
+* Yul: Introduce builtin `slotnum()` for the `SLOTNUM` opcode, available since the Amsterdam EVM version ([EIP-7843](https://eips.ethereum.org/EIPS/eip-7843)).
+
+### Compiler Features
+
+* Commandline Interface: Remove support for the experimental Language Server Protocol (LSP) mode.
+* EVM: Deprecate support for "constantinople", "petersburg", "istanbul" and "berlin" EVM versions.
+* Evmasm Optimizer: Improve performance of block deduplicator.
+* General: Improve performance throughout the compiler using Boost's flat versions of unordered set and map.
+* General: Remove support for the experimental Generic Solidity prototype (`pragma experimental solidity`).
+* General: Replace the current greedy stack shuffler in the experimental SSA CFG pipeline with a planning one.
+* SMTChecker: Emit a deprecation warning for the BMC engine.
+* SMTChecker: Support `block.slotnum`.
+* Yul Optimizer: `LoopInvariantCodeMotion` can now move expressions depending on function parameters out of loops.
+* Yul Optimizer: `UnusedStoreEliminator` can now recognize redundant memory and storage operations whose start offset or length is a function parameter.
+* Yul Optimizer: Remove the ineffective elimination of unused `returndatacopy()` operations in simple cases from UnusedStoreEliminator.
+
+### Bugfixes
+
+* Code Generator: Fix constants read in both checked and `unchecked` contexts within one contract getting the checked/unchecked semantics of whichever read was generated first via IR, which could remove a required overflow panic or introduce a spurious one.
+* Code Generator: Fix ICE on parenthesized custom error construction in require statement.
+* Code Generator: Fix uninitialized internal function pointers being read from a packed storage slot with the wrong value when a subsequent variable in the slot holds a non-zero value.
+* Commandline Interface: Reject `--model-checker-ext-calls` in unsupported input modes instead of silently ignoring it.
+* Commandline Interface: Report proper error instead of ICE on non-hex mixed-case address value given via `--libraries`.
+* Standard JSON Interface: Fix the entire output being replaced by a `JSONError` ("Error writing output JSON.") when an error message quotes a long source line and truncating it splits a multi-byte character.
+* Type Checker: Report an unimplemented feature error instead of ICE when a variable of a fixed point type is accessed in inline assembly.
+* Yul IR Code Generation: Encode custom error named parameters in declaration order instead of call-site order when used from a `require` function.
+* Yul Optimizer: Fix incorrect removal of `returndatacopy()` operations referencing a stale result of `returndatasize()`.
+
+### Build System
+
+* Update minimum version requirement of Boost to 1.83.0 for Windows build. This matches the minimum version for other systems.
+
+## How to Install/Upgrade?
+
+To upgrade to the new version of the Solidity Compiler, please follow the [installation instructions](https://docs.soliditylang.org/en/v0.8.37/installing-solidity.html) available in our documentation.
+You can download the release from GitHub: [v0.8.37](https://github.com/argotorg/solidity/releases/tag/v0.8.37).
+
+If you want to build from the source code, do not use the source archives generated automatically by GitHub.
+Instead, use the [`solidity_0.8.37.tar.gz` source tarball](https://github.com/argotorg/solidity/releases/download/v0.8.37/solidity_0.8.37.tar.gz) or check out the `v0.8.37` tag via git.
+
+And last but not least, we would like to give a big thank you to all the contributors who helped make this release possible!
--- a/src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
+++ b/src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
@@ -88,6 +88,10 @@ From Amsterdam on, `slotnum` is also a reserved identifier in Yul.
 
 Two experimental features that were restricted to the [experimental mode](https://docs.soliditylang.org/en/v0.8.37/using-the-compiler.html#experimental-mode) in 0.8.35 are now removed entirely: the Language Server Protocol mode (`--lsp`) and the Generic Solidity prototype (`pragma experimental solidity`).
 
+The experimental `ABIEncoderV2` pragma now emits a deprecation warning.
+The pragma has been redundant since `pragma abicoder v2` was introduced in 0.7.4 and ABI coder v2 became the default in 0.8.0, but older `forge-std` versions that many projects depend on still use it.
+It will be removed in a later release.
+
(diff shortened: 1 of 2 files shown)


src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+When `delete` was applied to an element of a `bytes` array in memory, the evmasm pipeline wrote 32 zero bytes starting at that element instead of clearing only the element's byte.
+The extra zero bytes overwrote up to 31 bytes following the element.
+Those bytes may belong to later elements of the same array or, when the element sits near the end of the array, to whatever is allocated after it.
+The same applies to `string` values converted to `bytes`.
+The IR pipeline is not affected, and the analogous assignment `b[i] = 0` clears exactly one byte on both pipelines.
+All versions up to and including 0.8.36 are affected.
+
+The bug was reported by [shaheenfazim](https://github.com/shaheenfazim) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Memory Byte Array Element Delete Clears Whole Word Bug](https://blog.soliditylang.org/2026/09/10/memory-byte-array-element-delete-clears-whole-word-bug/)
```
cameel: Do relative links work for these? We're linking to a file from the same repo.

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+title: "Solidity 0.8.37 Release Announcement"
+date: "2026-09-10"
+author: Solidity Team
+category: Releases
+---
+
+We have released [Solidity v0.8.37](https://github.com/argotorg/solidity/releases/tag/v0.8.37).
+
+This is primarily a bugfix release.
+It contains two important bugfixes, both rated low/medium severity, as well as fixes for several further bugs of lower severity, three of which are described below.
```
cameel: The important distinction here is that the smaller bugs are still codegen errors, so more problematic than e.g. plain ICEs but not security-relevant.

Maybe this would be clearer?
```suggestion
It contains two important bugfixes, both rated low/medium severity, as well as fixes for several smaller miscompilations that are not security issues, three of which are described below.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+
+Write-up: [Memory Byte Array Element Delete Clears Whole Word Bug](https://blog.soliditylang.org/2026/09/10/memory-byte-array-element-delete-clears-whole-word-bug/)
+
+### Spill Slot Collision Across Mutual Recursion
+
+The stack-to-memory mover of the IR pipeline works around the EVM's stack depth limit by relocating some local variables to fixed memory offsets, called spill slots.
+When the call graph contained mutual recursion, too few slots were reserved, so variables of two different functions could be assigned the same slot even though both could be live at the same time.
+One variable was then silently overwritten by the other.
+The bug is closely related to, but distinct from, the [Unsound Spill In Mutual Recursion Bug](https://blog.soliditylang.org/2026/07/09/unsound-spill-in-mutual-recursion-bug/) fixed in 0.8.36.
+That fix does not cover this case.
```
clonker:
```suggestion
```
the other one is already fixed and it's distinct, i would personally drop this sentence

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+Versions 0.7.2 through 0.8.36 are affected when compiling with `--via-ir`.
+
+The bug was reported by [Ng Sze Hon](https://github.com/gzeoneth) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Spill Slot Collision Across Mutual Recursion Bug](https://blog.soliditylang.org/2026/09/10/spill-slot-collision-across-mutual-recursion-bug/)
+
+## Other Bugfixes
+
+### Misordered Named Parameters in require with Custom Errors
```
cameel, line 42: Wait, this is also a security fix, it should be under "Important Bugfixes". It's not an ordinary bug, it's just low severity.

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+
+In the IR pipeline, the arguments of a custom error constructed with named arguments and passed to `require`, as in `require(condition, MyError({b: 2, a: 1}))`, were ABI-encoded in call-site order rather than declaration order.
```
clonker:
```suggestion
In the IR pipeline, the arguments of a custom error constructed with named arguments and passed to `require`, as in `require(condition, MyError({b: 2, a: 1}))`, were put on the stack in call-site order rather than declaration order, potentially leading to a wrong ABI-encoding thereof.
```

Perhaps this is better? I'd be happy to hear @cameel s thoughts on this though

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
cameel: Damn. I see that the PR added it under normal bugfixes in the changelog. I did not catch it and looks like no one else noticed. We need to move it.

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+In the IR pipeline, the code computing the value of a constant was generated once per contract, with either checked or unchecked arithmetic depending on which use site was generated first, and every other use site reused that code.
+For a constant whose initializer overflows its type, a read in a checked context could therefore miss its required overflow panic, or a read in an `unchecked` context could revert with a spurious one.
+
+## Notable Features and Changes
+
+### SSA CFG Planning Stack Shuffler (Experimental)
+
+The [experimental SSA CFG code generator](https://soliditylang.org/blog/2026/04/29/solidity-0.8.35-release-announcement/#experimental-ssa-cfg-code-generator) precomputes a stack layout for every point in the program.
+Its stack shuffler emits the `SWAP`, `DUP`, `POP` and `PUSH` instructions that get the stack from one layout to the next.
+Until now the shuffler worked greedily, applying one local repair step at a time until the target layout was reached.
```
clonker:
```suggestion
Until now the shuffler worked greedily, applying one local step at a time until the target layout was reached.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+All versions up to and including 0.8.36 are affected.
+
+The bug was reported by [shaheenfazim](https://github.com/shaheenfazim) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Memory Byte Array Element Delete Clears Whole Word Bug](https://blog.soliditylang.org/2026/09/10/memory-byte-array-element-delete-clears-whole-word-bug/)
+
+### Spill Slot Collision Across Mutual Recursion
+
+The stack-to-memory mover of the IR pipeline works around the EVM's stack depth limit by relocating some local variables to fixed memory offsets, called spill slots.
+When the call graph contained mutual recursion, too few slots were reserved, so variables of two different functions could be assigned the same slot even though both could be live at the same time.
```
cameel: This makes it sound like the recursion is enough to cause it that. That makes it seem more severe than it is. We should qualify it.

```suggestion
In certain call graph configurations containing mutual recursion too few slots were reserved, so variables of two different functions could be assigned the same slot even though both could be live at the same time.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,144 @@
...
+## Notable Features and Changes
+
+### SSA CFG Planning Stack Shuffler (Experimental)
+
+The [experimental SSA CFG code generator](https://soliditylang.org/blog/2026/04/29/solidity-0.8.35-release-announcement/#experimental-ssa-cfg-code-generator) precomputes a stack layout for every point in the program.
+Its stack shuffler emits the `SWAP`, `DUP`, `POP` and `PUSH` instructions that get the stack from one layout to the next.
+Until now the shuffler worked greedily, applying one local repair step at a time until the target layout was reached.
+0.8.37 replaces it with a planning shuffler.
+The new shuffler first computes an explicit plan that maps the items on the current stack to their positions in the target layout, and then realizes that plan bottom-up.
+If it gets stuck because an item is out of reach for `SWAP` or `DUP`, it drops the items above the blocking one and starts over from a new plan.
```
clonker:
```suggestion
If it gets stuck because an item is out of reach for `SWAP` or `DUP`, it compresses the stack to achieve reachability and starts over from a new plan.
```

clonker (commented):
some comments!

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
cameel: Yeah, I was just in the middle of the reviewing this and was about to suggest a change here :) Not done yet, but submitted my partial review to be able to respond to this.

The basis for this is my suggestion for updating the bug list entry from https://github.com/argotorg/solidity/pull/16453#discussion_r3973352915

My suggestion would be this:
```suggestion
In the IR pipeline, the arguments of a custom error constructed with named arguments and passed to `require`, as in `require(condition, MyError({b: 2, a: 1}))`, were put on the stack in call-site order rather than declaration order, potentially leading them being matched with wrong parameters during ABI encoding, possibly being misinterpreted as representing completely different types.
The encoder would produce a structurally valid payload matching the error signature but not reflecting the values used to instantiate the error.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,147 @@
...
+When the call graph contained mutual recursion, too few slots were reserved, so variables of two different functions could be assigned the same slot even though both could be live at the same time.
+One variable was then silently overwritten by the other.
+The bug is closely related to, but distinct from, the [Unsound Spill In Mutual Recursion Bug](https://blog.soliditylang.org/2026/07/09/unsound-spill-in-mutual-recursion-bug/) fixed in 0.8.36.
+Versions 0.7.2 through 0.8.36 are affected when compiling with `--via-ir`.
+
+The bug was reported by [Ng Sze Hon](https://github.com/gzeoneth) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Spill Slot Collision Across Mutual Recursion Bug](https://blog.soliditylang.org/2026/09/10/spill-slot-collision-across-mutual-recursion-bug/)
+
+## Other Bugfixes
```
cameel:
```suggestion
## Other Miscompilations
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,147 @@
...
+Only the revert data of a transaction that is already reverting is affected, so the bug has no impact on contract state.
+
+The bug was reported by Carl from [Spearbit](https://spearbit.com/) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Misordered Named Parameters in require with Custom Errors Bug](https://blog.soliditylang.org/2026/09/10/misordered-named-parameters-in-require-with-custom-errors-bug/)
+
+### Uninitialized Function Pointers in Packed Storage
+
+In the evmasm pipeline, reading an uninitialized internal function pointer from a packed storage slot yielded the wrong value when a variable stored after it in the same slot held a non-zero value.
+As a consequence, an equality comparison between such a pointer and a default-initialized local function pointer could be false, even though neither had ever been assigned.
```
cameel, line 59: I think that for these it would be useful to include a short summary of our reasoning that led us to not classify it as a security issue. Without that they may sound scarier than they actually are.

For reference, here's a summary I posted earlier in our initial analysis. That's too long, so I'm not saying to use all of it, but perhaps we could distill it into 2-3 sentences containing the most important facts. 

> - Summary: valid ordinary bug but not a security issue because the triggers are very specific, likelihood is low and you have to intentionally ignore all the warning signs to run into it.
> - Analysis: what we determined about the bug
>     - Reproducible (Evmasm only)
>     - Does not affect pointers not packed with other values.
>     - Does not affect pointers that are assigned a value before the first comparison.
>     - Affects assignments (but it does not make the impact bigger because ...)
>     - Does not affect function calls (unless it happens to hit a JUMPDEST by chance)
> - Impact:
>     - Internal function pointer comparisons already have many caveats, which makes relying on their equality a bad practice
>         - Fragile due to dependence on bytecode
>         - The optimizer may break equality
>         - Putting them in storage is questionable and requires careful precautions to be safe
>     - The compiler does warn about it and it is documented.
>     - In the next breaking version we are even going to disallow both comparisons and storage variables of these types
>     - Triggering it depends on very specific usage patterns we do not think are likely in practice
>     - Reliably detectable through unit testing

Especially the fact that it only affects evmasm is very relevant. And I'd link to the docs: https://docs.soliditylang.org/en/v0.8.37/types.html#value-stability-across-contract-updates

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Constants Read in Both Checked and Unchecked Contexts
+
+In the IR pipeline, the code computing the value of a constant was generated once per contract, with either checked or unchecked arithmetic depending on which use site was generated first, and every other use site reused that code.
+For a constant whose initializer overflows its type, a read in a checked context could therefore miss its required overflow panic, or a read in an `unchecked` context could revert with a spurious one.
```
cameel, line 69: And here's the relevant bit of our response for this. Again a bit too long, but could be reworded:

> As stated above, we confirm the miscompilation but we do not consider it a security issue and classify it as an ordinary bug.
> All operands of the affected operation are compile-time constants, so the miscompiled read is deterministic and independent of any transaction input, and no attacker can influence it.
> For constants, the checked arithmetic acts as a sanity check rather than as validation, because the result cannot change based on input.
> Exploitation would therefore depend on code that intentionally relies on a constant that always reverts at runtime, or on one that is meant to work inside `unchecked` blocks but not outside of them, neither of which we see as plausible intentional design.
> The realistic failure mode is a mistake in a contract with insufficient test coverage: the spurious panic direction fails on the first execution of the affected function, and the dropped-check direction yields a fixed wrapped value, both of which testing would easily surface.
> Reaching the defect additionally requires a source that reads (possibly transitively via a constant referencing it) one overflowing typed constant from both a checked and an `unchecked` context within one contract's object, which we expect to be rare in practice.

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,147 @@
...
+0.8.37 replaces it with a planning shuffler.
+The new shuffler first computes an explicit plan that maps the items on the current stack to their positions in the target layout, and then realizes that plan bottom-up.
+If it gets stuck because an item is out of reach for `SWAP` or `DUP`, it compresses the stack to achieve reachability and starts over from a new plan.
+The more structured approach is easier to maintain, and initial benchmarks show it to be computationally more efficient than the greedy one while producing bytecode of roughly the same size.
+
+The SSA CFG code generator remains experimental and is enabled with `--experimental --via-ssa-cfg`.
+
+### Support for the SLOTNUM Opcode
+
+The Amsterdam EVM version, supported since 0.8.36, introduces the `SLOTNUM` opcode ([EIP-7843](https://eips.ethereum.org/EIPS/eip-7843)), which returns the beacon chain slot number of the current block.
```
cameel: Should we clarify that the support is still experimental and incomplete? It makes it sound as if it's fully supported from 0.8.36.

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,147 @@
...
+0.8.37 exposes it in Solidity as `block.slotnum` of type `uint64`, and in Yul and inline assembly as the builtin `slotnum()`.
```
cameel: We could link to the list of builtins here: https://docs.soliditylang.org/en/v0.8.37/units-and-global-variables.html#block-and-transaction-properties

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,147 @@
...
+The SMTChecker supports the new member as well.
+`block.slotnum` and `slotnum()` are available only when compiling with `--evm-version amsterdam` or a later EVM version.
+From Amsterdam on, `slotnum` is also a reserved identifier in Yul.
+
+### Removals and Deprecations
+
+Two experimental features that were restricted to the [experimental mode](https://docs.soliditylang.org/en/v0.8.37/using-the-compiler.html#experimental-mode) in 0.8.35 are now removed entirely: the Language Server Protocol mode (`--lsp`) and the Generic Solidity prototype (`pragma experimental solidity`).
```
cameel: We should mention why we're removing them:
- The LSP was unfinished and unmaintained. We're not abandoning plans for a LSP but keeping such tools so tightly integrated with the compiler does not mesh with our plans to make the compiler leaner and more maintainable. They should be standalone.
- Generic Solidity was an early prototype of Core Solidity. It was an attempt to refactor the compiler in place and iterate on the next version of the language in the same codebase. This was neither fast nor flexible enough and we abandoned it in favor of starting standalone prototype (solcore) that's better suited for experimentation and will be a base for a future production implementation.

nikola-matic:
```suggestion
Two features that were restricted to the [experimental mode](https://docs.soliditylang.org/en/v0.8.37/using-the-compiler.html#experimental-mode) in 0.8.35 are now removed entirely: the Language Server Protocol mode (`--lsp`) and the Generic Solidity prototype (`pragma experimental solidity`).
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,147 @@
...
+
+The experimental `ABIEncoderV2` pragma now emits a deprecation warning.
+The pragma has been redundant since `pragma abicoder v2` was introduced in 0.7.4 and ABI coder v2 became the default in 0.8.0, but older `forge-std` versions that many projects depend on still use it.
+It will be removed in a later release.
```
cameel: This should explain why we're doing it this way. I.e. we consider the removal non-breaking and will remove the pragma way before 0.9, but we wanted to give projects a chance to drop it first to prevent sudden and completely avoidable breakage.

How about this?
```suggestion
The pragma has been redundant since `pragma abicoder v2` was introduced in 0.7.4 and ABI coder v2 became the default in 0.8.0.
It is a leftover from back when ABI coder v2 was still experimental and as such its removal does not require a breaking version.
However, since many projects still seem to use it (often indirectly, through a dependency on older `forge-std` versions), we decided to give them a nudge before we proceed with the removal.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
czepluch: yeah, fixed throughout

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+title: "Solidity 0.8.37 Release Announcement"
+date: "2026-09-10"
+author: Solidity Team
+category: Releases
+---
+
+We have released [Solidity v0.8.37](https://github.com/argotorg/solidity/releases/tag/v0.8.37).
+
+This is primarily a bugfix release.
+It contains three important bugfixes, two rated low/medium severity and one very low, as well as fixes for several smaller miscompilations that are not security issues, two of which are described below.
```
cameel, line 11: I mean, after my previous suggestion this needs further adjustment - we have several bugfixes but only two of those were miscopmilations.

```suggestion
It contains three important bugfixes, two rated low/medium severity and one very low, as well as fixes for two smaller miscompilations that are not security issues.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+The bug was reported by [Ng Sze Hon](https://github.com/gzeoneth) through the Ethereum Foundation's bug bounty program.
+
+Write-up: [Spill Slot Collision Across Mutual Recursion Bug](/blog/2026/09/10/spill-slot-collision-across-mutual-recursion-bug/)
+
+### Misordered Named Parameters in require with Custom Errors
+
+In the IR pipeline, the arguments of a custom error constructed with named arguments and passed to `require`, as in `require(condition, MyError({b: 2, a: 1}))`, were put on the stack in call-site order rather than declaration order.
+During ABI encoding they could therefore be matched with the wrong parameters and be misinterpreted as values of completely different types.
+The encoder would produce a structurally valid payload matching the error signature but not reflecting the values used to instantiate the error.
+Plain `revert MyError({...})` statements are not affected.
```
cameel, line 47:
```suggestion
Plain `revert MyError({...})` statements or any other language construct that allows named parameters are not affected.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Other Miscompilations
+
+### Uninitialized Function Pointers in Packed Storage
+
+In the evmasm pipeline, reading an uninitialized internal function pointer from a packed storage slot yielded the wrong value when a variable stored after it in the same slot held a non-zero value.
+As a consequence, an equality comparison between such a pointer and a default-initialized local function pointer could be false, even though neither had ever been assigned.
+
+We do not classify this as a security issue.
+It only affects the evmasm pipeline, only pointers packed with other values in the same slot and only pointers that are compared before ever being assigned, and calls through the pointer are not affected.
```
cameel, line 62:
```suggestion
The bug only affects the evmasm pipeline and has no effect on function calls made through such pointers.
Only the very specific case of packed pointers being compared before ever being assigned produces wrong results.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+Relying on the equality of internal function pointers is fragile in any case, since their values depend on the exact bytecode and are [not stable across contract updates](https://docs.soliditylang.org/en/v0.8.37/types.html#value-stability-across-contract-updates), which is why the compiler already warns about such comparisons and we plan to disallow them in the next breaking release.
```
cameel, line 63: Let's make this sound a bit less casual:

```suggestion
Relying on the equality of internal function pointers is fragile, since their values depend on the exact bytecode and are [not stable across contract updates](https://docs.soliditylang.org/en/v0.8.37/types.html#value-stability-across-contract-updates), which is why the compiler already warns about such comparisons and we plan to disallow them in the next breaking release.
```

clonker (approved):

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+Triggering the bug requires a very specific usage pattern, and it is reliably detectable through unit testing.
+
+### Constants Read in Both Checked and Unchecked Contexts
+
+In the IR pipeline, the code computing the value of a constant was generated once per contract, with either checked or unchecked arithmetic depending on which use site was generated first, and every other use site reused that code.
+For a constant whose initializer overflows its type, a read in a checked context could therefore miss its required overflow panic, or a read in an `unchecked` context could revert with a spurious one.
+
+We do not classify this as a security issue either.
```
cameel, line 71: I'd drop this.
```suggestion
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+All operands of the affected read are compile-time constants, so the miscompiled result is deterministic, independent of any transaction input and outside of an attacker's influence.
+Exploiting it would require code that intentionally relies on a constant that always reverts at runtime, or on one that is meant to work inside `unchecked` blocks but not outside of them, neither of which is a plausible design.
+The realistic failure mode is a mistake in a contract with insufficient test coverage: the spurious panic fails on the first execution of the affected function, and the missing check yields a fixed wrapped value, both of which testing would surface.
```
cameel, line 74: This is a bit dense so I'd reword. Also, while not incorrect, this usage of "surface" as a verb somehow started becoming very common recently, but it always sounds kinda awkward to me.

```suggestion
The realistic failure mode is a mistake in a contract with insufficient test coverage: the evaluation of the constant reverts with a `Panic` on the first execution of the affected function, and the missing check yields a fixed wrapped value, both of which would be easily discoverable in testing.
```

src/posts/2026-09-10-solidity-0.8.37-release-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+
+The SSA CFG code generator remains experimental and is enabled with `--experimental --via-ssa-cfg`.
+
+### Support for the SLOTNUM Opcode
+
+The Amsterdam EVM version introduces the `SLOTNUM` opcode ([EIP-7843](https://eips.ethereum.org/EIPS/eip-7843)), which returns the beacon chain slot number of the current block.
+0.8.37 exposes it in Solidity as [`block.slotnum`](https://docs.soliditylang.org/en/v0.8.37/units-and-global-variables.html#block-and-transaction-properties) of type `uint64`, and in Yul and inline assembly as the builtin `slotnum()`.
+The SMTChecker supports the new member as well.
+`block.slotnum` and `slotnum()` are available only when compiling with `--evm-version amsterdam`.
+Support for Amsterdam was added in 0.8.36 and is still experimental and incomplete.
```
cameel, line 97:
```suggestion
This is another step towards adding support for all Amsterdam features, which started with 0.8.36 and is still experimental.
```

cameel (commented):
The content is fine now, but it could still use a few smaller tweaks. See my suggestions below.

